PR #1, opened 2024-08-02
Feat: Initialize events

hetd54:
This PR adds the data needed to create the Calendar Events section of the index page. I didn't want this PR to get TOO large, so there isn't much reactivity here or a lot of design work. Those will be handled in another PR.

## Components

Two calendar components -- the Calendar container and the event container

The Calendar.tsx component fetches data from the events api, to be used to pass props to children. Toggle buttons for Monthly, Weekly, and Upcoming exist, but their state does not currently affect the display.

CalenderEvent.tsx component is the "Upcoming" display (and currently the only display); might refactor to "UpcomingEvent.tsx" if the component isn't going to be used elsewhere.

## Pages

page.tsx is the index page and currently only displays the calendar section.

pages/about.tsx is the about page


src/hooks/events.js
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from "react";
+import { getStringDate } from "../utils";
+
+export function getEventData() {
```
anna-murphy: I think this needs to be `useGetEventData` or something else with prefixed with `use` but I like this a lot! I had commented on this before I saw the refactor and this is really good.

src/utils.js
```diff
@@ -0,0 +1,68 @@
...
+ * Generates the string representation of a date.
+ *
+ * @param {number} month The integer representation of the month.
+ * @param {number} day The integer representation of the day of the month.
+ * @param {number} year The integer representation of the year.
+ * @returns {string} The date as a string.
+ */
+export const getStringDate = (month, day, year) => {
+  const mm = month < 10 ? "0" + month : month;
+  const dd = day < 10 ? "0" + day : day;
```
anna-murphy: This looks like a great use case for [`padStart`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)!

```suggestion
  const mm = month.toString().padStart(2, "0");
  const dd = day.toString().padStart(2, "0");
```
 I think this should work....

src/utils.js
anna-murphy, line 1: Why *not* make this a `.ts` file? When writing utility functions especially, having an extra layer of type safety can be useful for me in the dev process.

src/utils.js
```diff
@@ -0,0 +1,68 @@
...
+//
+
+/**
+ * Calculates the number of weeks in the current month.
+ *
+ * @param {number} month The month.
+ * @param {number} year The year.
+ * @returns {number} The number of weeks in the month.
+ */
+export const weeksInMonth = (month, year) => {
```
anna-murphy: Any reason why you're writing these as anonymous functions and not as named functions (eg `export function weeksInMonth(....`)?

src/utils.js
```diff
@@ -0,0 +1,68 @@
...
+  "Thurs",
+  "Fri",
+  "Sat",
+];
+
+export const DISPLAY_YEAR = new Date().getFullYear();
+
+export const DISPLAY_MONTH = new Date().getMonth() + 1;
+
+export const DISPLAY_DAY = new Date().getDate();
```
anna-murphy: Are these initialized during build time? If so, they would get out of sync with the current time pretty quick, right? It might be worth making these functions?

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,75 @@
...
+            Upcoming
+          </p>
+        </div>
+      </div>
+
+      <div className="flex flex-wrap justify-between gap-4">
+        {data &&
+          data.map((e: DataProps, i) => {
+            return (
+              <div key={i}>
```
anna-murphy: You're not technically supposed to use the index of a thing as it's key. Could you do something like this instead?

```suggestion
              <div key={`${e.date_time}-${e.contact_info}-${e.title}`}>
```

src/components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,75 @@
...
+                <CalendarEvent
+                  date_time={e.date_time}
+                  title={e.title}
+                  url={e.url}
+                  date={e.date}
+                  date_utc={e.date_utc}
+                  description_long={e.description_long}
+                  contact_info={e.contact_info}
+                  description={e.description}
```
anna-murphy: This might be silly, but if you're just passing all the values from the object like this, I wonder if it would make for a better API to have `CalendarEvent` only have one prop `eventData` that just takes the value from the data list? Just a thought though.

anna-murphy (approved):
Looks pretty good! Any chance you're going to get live Firebase previews in sometime? The code looks fine but I haven't run it locally so idk how it looks. (That would be a different PR I'm sure, I'm more asking about what's on your plate and how you're thinking about approaching things.)

src/utils.js
hetd54: tbh I just copied the utils.js file over from the nuxt ccv website repo; I can refine everything though!

src/components/calendar/Calendar.tsx
hetd54: How come? Is it because if you have more than one map the keys can be the same within the view?

src/utils.js
hetd54: Yum, thanks for identifying a new (for me) method!

src/components/calendar/Calendar.tsx
hetd54: the eventData object is HUGE -- does passing the whole thing make the child component take longer to load?

anna-murphy: I don't... *think* so? My guess is that once it's loaded into memory it's just passed by reference. If not, is there a way to trim down the object you get from the API call?

src/components/calendar/Calendar.tsx
anna-murphy: React uses the key of a fragment to cache the rendered output for the result of the map, and I'm pretty sure indexes aren't guaranteed to be the same (if you sort an array, the keys would no longer match up). Like, it works, but there's a performance hit. 

Here's a SO post on this if you're curious to read more! https://stackoverflow.com/a/59518082

src/utils.js
anna-murphy: You do what you want! I prefer the function syntax but if you prefer the anonymous function, it's all good.

src/utils.js
hetd54: Looked into this, and this isn't even used in the current site. I'm removing it.

hooks/events.js
anna-murphy, line 1: ❤️

package.json
```diff
@@ -19,10 +19,11 @@
     "@types/node": "^20",
     "@types/react": "^18",
     "@types/react-dom": "^18",
+    "autoprefixer": "^10.4.20",
     "eslint": "^8",
     "eslint-config-next": "14.2.5",
-    "postcss": "^8",
-    "tailwindcss": "^3.4.1",
+    "postcss": "^8.4.41",
```
anna-murphy: Probably not for this PR, but what does postcss do here? do we need it still?

components/calendar/CalendarEvent.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <div className="bg-gray flex gap-6 overflow-hidden column min-w-[30ch]">
+      <div className="event">
+        <div className="text-neutral-900 text-sm">{normalDate}</div>
+        <div>
+          <a
+            className="text-xl font-bold text-secondary-blue-500 no-underline leading-4"
+            href={data.url}
+            target="_blank"
+          >
+            {" "}
```
anna-murphy: Do you need the `{" "}`'s? I can't tell from reading the CSS, but you might be able to get away with some padding? Are the title and time supposed to be in the same line?

components/calendar/Calendar.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </p>
+        </div>
+      </div>
+
+      <div className="flex flex-wrap justify-between gap-4">
+        {data &&
+          data.map((event: DataProps) => {
+            return (
+              <div
+                key={`${event.date_utc}-${event.contact_info}-${event.title}`}
```
anna-murphy, line 58: Can you put the key on the `CalendarEvent` and remove the `div`?

anna-murphy (commented):
Looks pretty good!~

utils.ts
eldu, line 1: Were these utils grabbed from the previous website? Just curious

utils.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns {string} The date as a string.
+ */
+export function getStringDate(
+  month: number,
+  day: number,
+  year: number,
+): string {
+  const mm = month.toString().padStart(2, "0");
+  const dd = day.toString().padStart(2, "0");
+  return year + "-" + mm + "-" + dd;
```
eldu: For formatting strings, I'd probably lean to template strings nowadays. Concatenation does work  though! So no need to change -- https://www.w3schools.com/js/js_string_templates.asp

```suggestion
  return `${year}-${mm}-${dd}`;
```

eldu (commented):
throwing in some quick comments. looks good overall !

utils.ts
hetd54: Yes, grabbed directly from the old site! Made some minor changes from some of Anna's comments, and I'll address yours as well :)

components/calendar/Calendar.tsx
hetd54: Yes, but addressing this refactor in another PR -- Going to prob rename and move it to an EventsSection component

package.json
hetd54: hmm, looks like no, removed it

eldu (approved):

package.json
anna-murphy: 🤩

anna-murphy (approved):

PR #2, opened 2024-08-08
chore: setup firebase

hetd54:


eldu (commented):
More so a question

When I click on the preview link, it's blank which is like how the `/pages/index.tsx` looks however in the `public/index.html` the html has a div that says "Firebase Hosting Setup Complete"

Is there supposed to be a `public/index.html` file? I feel like I usually just see favicons.

hetd54:
> More so a question
> 
> When I click on the preview link, it's blank which is like how the `/pages/index.tsx` looks however in the `public/index.html` the html has a div that says "Firebase Hosting Setup Complete"
> 
> Is there supposed to be a `public/index.html` file? I feel like I usually just see favicons.

Good catch -- this was created when I have the public folder as the build folder erroneously. Removed!

eldu (approved):

PR #3, opened 2024-08-08
Chore: add prettier and eslint

hetd54:


.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "trailingComma": "es5",
+  "semi": false,
```
anna-murphy, line 3: 🪦 my beloved semicolons... 🪦

app/layout.tsx
```diff
@@ -1,23 +1,23 @@
...
   title: "CCV",
   description: "Center for Computation & Visualization",
-};
+}
 
 export default function RootLayout({
   children,
 }: Readonly<{
-  children: React.ReactNode;
+  children: React.ReactNode
```
anna-murphy, line 16: Probably not for this PR, but I like to type things like this as `React.propsWithChildren`. That's technically a type helper and will add the `children` type to whatever you pass into it (`React.propsWithChildren<MyComponentProps>`), but can be used just a a standalone type. Just my preference though!

anna-murphy (approved):
Always a chore to add the linter... looks good though! any reason you didn't want semicolons / commas? i like the verbosity of them personally but it's totally okay to skip them if that feels better to you

PR #15, opened 2024-08-14
Feat buttons

### LINKED ISSUE #6: Button Component
https://github.com/brown-ccv/ccv-website-nuxt/blob/main/components/base/DButton.vue

hetd54:
This PR seeks to add three button functions within the Button.tsx component:

- Button: blue background, the most common button used on pages
- ButtonSecondary: yellow background, used in headers and sorting/filtering
- ButtonTertiary: offwhite background, used in the navbar

### DIFF
--- a/components/Button.tsx
+++ b/components/Button.tsx
@@ -1,0 +1,47 @@
+import React, { type ReactNode } from "react";
+
+interface ButtonProps
+  extends React.PropsWithChildren<
+    React.ButtonHTMLAttributes<HTMLButtonElement>
+  > {
+  icon?: ReactNode;
+}
+
+const buttonStyles =
+  "flex items-center gap-2 px-3 py-2 w-max rounded transition-all hover:shadow-md";
+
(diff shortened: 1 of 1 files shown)


components/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+      className={`${buttonStyles} bg-secondary-blue-500 text-white hover:bg-secondary-blue-700`}
+      {...delegated}
+    >
+      {icon}
+      <span>{children}</span>
+    </button>
+  );
+}
+
+export function ButtonSecondary({ icon, children, ...delegated }: ButtonProps) {
```
anna-murphy: This might just be a me thing, but I feel like I'd prefer an API thats a single `Button` component with a `variant` prop that's optional and is either `"secondary"` or `"tertiary"`. Do you prefer three different components?

anna-murphy (approved):
Nice and simple! I don't see any buttons on the preview... Are they used anywhere and I'm just bad at looking?

anna-murphy (approved):
🌠

PR #17, opened 2024-10-09
Feat: Calendar weekly event views

hetd54:
This PR adds the "weekly" view for the CCV Events section.

# Changes

The old page looks like:

Screenshot 2024-10-09 at 11 35 44 AM

New page looks like:
Screenshot 2024-10-15 at 4 52 44 PM


components/calendar/CalendarWeekly.tsx
```diff
@@ -0,0 +1,410 @@
...
+                  </div>
+                </div>
+                <div />
+                <div>
+                  <div className="sticky left-0 z-20 -ml-14 -mt-2.5 w-14 pr-2 text-right text-xs leading-5 text-gray-400">
+                    11PM
+                  </div>
+                </div>
+                <div />
+              </div>
```
anna-murphy: I'd _love_ to wrap these in some kind of a map.... it looks like all the styling is pretty similar, so you might be able to do it like:

```jsx
const TIMES = ["1AM", "2AM", /* etc */ ];
function CalendarWeekly() {
  return (
    <>
      {/* Whatever else goes in here */}
      {TIMES.map((time) => (
        <div>
          <div className="sticky left-0 z-20 -ml-14 -mt-2.5 w-14 pr-2 text-right text-xs leading-5 text-gray-400">
            {time}
          </div>
        </div>
      ))}
    </>
  );
}
```

Should reduce a lot of the repeated markup, which would make it easier to read.

components/calendar/CalendarWeekly.tsx
```diff
@@ -0,0 +1,410 @@
...
+                <div
+                  className={`col-start-6 row-span-full ${todayRow === 5 && isSameDay(activeDate, currentDate) ? "bg-secondary-yellow-100" : ""}`}
+                />
+                <div
+                  className={`col-start-7 row-span-full ${todayRow === 6 && isSameDay(activeDate, currentDate) ? "bg-secondary-yellow-100" : ""}`}
+                />
+                <div
+                  className={`col-start-8 row-span-full w-8 ${todayRow === 7 && isSameDay(activeDate, currentDate) ? "bg-secondary-yellow-100" : ""}`}
+                />
+              </div>
```
anna-murphy: Same with these. I think this would require more supporting data to work, but could be good to simplify.

utils.ts
anna-murphy, line 1: This might be too silly, but if this is utils only for `CalendarWeekly` I wonder if it'd be better to move it to the `calendar/` folder? I feel like having a structure of component, associated hooks, and utility functions in the same folder makes sense to me.

components/calendar/CalendarWeekly.tsx
anna-murphy: Update: you mentioned that you might not be able to do this, but I'm happy to take a look with you about it if you'd like~

components/calendar/CalendarWeekly.tsx
```diff
@@ -0,0 +1,410 @@
...
+        new Date(yearEvent, monthEvent, dateEvent)
+      )
+
+      const calColor =
+        event.title === "CCV Office Hours"
+          ? "bg-primary-50 hover:bg-primary-500"
+          : event.title ===
+              "COBRE CBC Computational Biology Walk-in Office hours"
+            ? "bg-secondary-blue-50 hover:bg-secondary-blue-300"
+            : "bg-secondary-yellow-50 hover:bg-secondary-yellow-100"
```
anna-murphy: Do you get Tailwind Autocomplete here? This is something I've sometimes struggled with in my projects.

components/calendar/CalendarWeekly.tsx
anna-murphy, line 1: Something I'm curious about is whether or not you think it might be worthwhile to constrain the time of day on the calendars? Like, we're rarely going to be having events before 8:00 AM and after 9:00 PM or whatever. Does it make sense to limit that do you think?

components/calendar/CalendarEvent.tsx
```diff
@@ -1,37 +1,37 @@
...
-            target="_blank"
-          >
-            {data.title}
-          </a>
-        </div>
-        <div className="text-primary-500 text-xs">{data.date_time}</div>
-        <div className="py-2">
-          {data.description_long && <p>{descriptionLong}</p>}
-        </div>
+    <div className="event">
```
anna-murphy, line 19: The combination of tailwind classed and like custom classes is interesting to me. When do you decide to use one over the other?

components/EventSection.tsx
```diff
@@ -0,0 +1,140 @@
...
+  useEffect(() => {
+    fetch(
+      "https://events.brown.edu/live/json/events/description_long/true/group/Center%20for%20Computation%20and%20Visualization%20%28CCV%29/start_date/" +
+        today +
+        "/"
+    )
+      .then((res) => res.json())
+      .then((data) => {
+        setDataFuture(data)
+        console.log(data)
```
anna-murphy:
```suggestion
```

Probably remove the console log.

components/EventSection.tsx
```diff
@@ -0,0 +1,140 @@
...
+            >
+              Monthly
+            </p>
+            <p
+              id="upcoming"
+              className={view === "upcoming" ? "selected" : ""}
+              onClick={() => setView("upcoming")}
+            >
+              Upcoming
+            </p>
```
anna-murphy: This is another thing that I would array map I think. Not sure if you think it makes sense, but it might be cleaner. 

Also, why `<p>`? I'm wondering if a `<button>` makes sense, or at least setting the `aria-role` to button so SR's note it as interactable.

anna-murphy (approved):
Looks really good! I have a few suggestions for cleaning stuff up, but the proverbial jury might still be out on whether or not you *can* `array.map` some of these things....

Also! I'd love to pick your brain sometime about working with Tailwind UI 🌠

components/calendar/CalendarWeekly.tsx
hetd54: @anna-murphy With this updated code, the cal events stack at the beginning of the week instead of at their proper locations.

```
  const [week, setWeek] = useState<Array<weekProps>>([])

  useEffect(() => {
    // Generate unique IDs for list items
    const updatedWeeks = Array.from(Array(8).keys()).map((_, i) => ({
      id: self.crypto.randomUUID(),
      num: i,
    }))
    setWeek(updatedWeeks)
  }, [])


                {week.map(function (week) {
                  return (
                    <div
                      key={week.id}
                      className={`row-span-full col-start-${week.num + 1} ${
                        week.num === 7 ? "w-8" : ""
                      } ${week.num === todayRow && isSameDay(activeDate, currentDate) ? "bg-secondary-yellow-100" : ""}`}
                    />
                  )
                })}
```

hetd54: Screenshot 2024-10-15 at 2 20 55 PM

components/calendar/CalendarWeekly.tsx
hetd54: No I don't -- but these are custom colors I specified in the config so it wasn't a biggie. I usually create the classes within the html first before abstracting to address this sadness

components/calendar/CalendarWeekly.tsx
hetd54: From conversations with Paul Stey, I think keeping the full cal is preferred, as we might advertise via our calendar a workshop that another dept puts on. On page load, the view gets set to the current time so that you don't start at midnight.

components/calendar/CalendarEvent.tsx
hetd54: If the combo of tailwind classes gets overly long or reused, I think it is better to have a custom class that folks can reference/change, otherwise it gets REALLY annoying to look at within the html.

utils.ts
hetd54: Yeah, I agree -- I am not even sure that other sections are going to need util functions

components/EventSection.tsx
```diff
@@ -0,0 +1,134 @@
...
+    currentDate.getFullYear()
+  )
+
+  const CAL_VIEW_ARRAY = ["Weekly", "Monthly", "Upcoming"]
+
+  useEffect(() => {
+    fetch(
+      "https://events.brown.edu/live/json/events/description_long/true/group/Center%20for%20Computation%20and%20Visualization%20%28CCV%29/start_date/" +
+        today +
+        "/"
```
eldu: Convert to a template string

```suggestion
      `https://events.brown.edu/live/json/events/description_long/true/group/Center%20for%20Computation%20and%20Visualization%20%28CCV%29/start_date/${today}/`
```

components/EventSection.tsx
```diff
@@ -0,0 +1,134 @@
...
+    )
+      .then((res) => res.json())
+      .then((data) => {
+        setDataFuture(data)
+        setLoading(false)
+      })
+    fetch(
+      "https://events.brown.edu/live/json/events/description_long/true/group/Center%20for%20Computation%20and%20Visualization%20%28CCV%29/start_date/-2 months" +
+        today +
+        "/"
```
eldu:
```suggestion
      `https://events.brown.edu/live/json/events/description_long/true/group/Center%20for%20Computation%20and%20Visualization%20%28CCV%29/start_date/-2 months${today}/`
```

components/EventSection.tsx
```diff
@@ -0,0 +1,134 @@
...
+import CalendarWeekly from "@/components/calendar/CalendarWeekly"
+
+export interface DataProps {
+  id: string
+  date_time: string
+  date: string
+  title: string
+  date_iso: string
+  date_utc: string
+  date2_utc: string
```
eldu, line 15: Why are there so many date fields? Can these be cleaned up?

Also rather than `date2` can we instead do something like `start_date` and `end_date`?

components/calendar/utils.ts
```diff
@@ -53,7 +39,7 @@ export const ALL_DAYS_OF_WEEK = [
   "Sun",
   "Mon",
   "Tues",
-  "Weds",
+  "Wed",
```
eldu, line 42: If Tuesday is Tues and Thursday is Thurs, then I think Wednesday can stay as `Weds`. Otherwise I'd change to Tue, Wed, Thur

eldu (changes requested):
I added in some comments, though let's talk about it in today's website meeting !

components/EventSection.tsx
hetd54: These are all fields from the data fetched using LiveWhale API: https://support.livewhale.com/live/blurbs/api#Authenticated-Requests

I need the iso date for accurate calculations (how many mins into the day is this event?), and the utc dates for displaying on the page

eldu (approved):

PR #18, opened 2024-10-15
Fix: Add JS Array Maps to Weekly Calendar View

anna-murphy:
I create the arrays that are mapped over in the JSX outside the component, but they can theoretically be moved inline -- not really sure if it makes a difference, but I think this is a little cleaner?


eldu (commented):
Ooh, I think this is great! I think you need to resolve some conflicts though

anna-murphy:
actually I think heather solved this a different way! i'll close the PR, sorry~

PR #20, opened 2024-10-22
Initialize footer

hetd54:


components/Footer.tsx
```diff
@@ -0,0 +1,159 @@
...
+            <div className="flex w-2/3 gap-3.5">
+              <LogoBrown />
+              <LogoCcv />
+            </div>
+          </div>
+
+          <div className="mt-8 grid grid-cols-1 lg:grid-cols-5 gap-8 xl:col-span-2">
+            {Object.entries(navigation).map(([_, contents]) => {
+              return (
+                <div key={_}>
```
anna-murphy: I personally don't love using the `_` here, it's a little nitpicky, but if you're going to use the value I think you should either change the `_` to `key` or something or use the `href` for the key value. So it could look like either of these:
 - `<div key={key> ...`
 - `<div key={href}> ...`

components/assets/LogoBrown.tsx
```diff
@@ -9,7 +13,7 @@ const LogoBrown: React.FC = () => {
         xmlns="http://www.w3.org/2000/svg"
         x="0px"
         y="0px"
-        width="100%"
+        width={width}
```
anna-murphy, line 16: love this pattern!

app/globals.css
```diff
@@ -24,18 +24,9 @@
     }
 
     body {
-        @apply grid;
-        @apply bg-neutral-50;
-        grid-template-columns:
-      1fr
-      min(1000px, 100%)
-      1fr;
+        @apply bg-white;
```
anna-murphy, line 27: what happened to the styles that got removed here? that's intentional?

components/Footer.tsx
```diff
@@ -0,0 +1,159 @@
...
+      <div className="mx-auto max-w-7xl px-6 pb-8 pt-8 sm:pt-24 lg:px-8 lg:pt-16">
+        <div className="xl:grid xl:grid-cols-2 xl:gap-8">
+          <div className="space-y-8">
+            <div className="flex w-2/3 gap-3.5">
+              <LogoBrown />
+              <LogoCcv />
+            </div>
+          </div>
+
+          <div className="mt-8 grid grid-cols-1 lg:grid-cols-5 gap-8 xl:col-span-2">
```
anna-murphy: hmm i was checking this out on the preview and maybe we should have a `md` style here too? like maybe `md:grid-cols-3` or something? it's a little awkward in the `sm` to `lg` transition period i think

anna-murphy (approved):
pretty good~ the only thing id flag is maybe a `md` style on the footer links, but other than that i think you're okay to merge 🌠

app/globals.css
hetd54: Yes! Made the whole page styling weirdly boxy so I removed

PR #21, opened 2024-11-20
Navbar subroutes

hetd54:
# Added by ShadCN:

UI folder and subsequent components
components.json
libs/utils.js
Edits to talwind config
Deps to Radix in the package.json


components/Navbar.tsx
```diff
@@ -0,0 +1,278 @@
...
+      </section>
+
+      {/* Navigation Menu for Desktop */}
+      <NavigationMenu.Root className="hidden lg:block">
+        <NavigationMenu.List className="flex list-none space-x-5">
+          {navigation.map((item) => {
+            return (
+              <NavigationMenu.Item key={item.href}>
+                <NavigationMenu.Trigger className="group inline-flex h-9 items-center justify-center rounded-md bg-background px-4 py-2 text-md font-medium transition-colors hover:bg-accent hover:text-accent-foreground focus:bg-accent focus:text-accent-foreground focus:outline-none disabled:pointer-events-none disabled:opacity-50 data-[active]:bg-accent/50 data-[state=open]:bg-accent/50">
+                  {item.title}{" "}
```
anna-murphy: What's your preference on `{" "}` versus margin / padding to get the same effect? I was of the feeling that text in the markup should be for the text only but i could be convinced otherwise!

components/Navbar.tsx
```diff
@@ -0,0 +1,278 @@
...
+        href: "/about#diversity-statement",
+      },
+    ],
+  },
+  {
+    title: "Blog",
+    href: "/blog",
+    routes: [],
+  },
+]
```
anna-murphy: This is a very Me thing, but I almost feel like I'd want this in a different file? Like in `components/Navbar` have `Navbar.tsx` and `NavItems.ts`. But given that your components folder is flat, I'm not sure it's worth it.... 🤔 

I do like this pattern! Im just trying to think if there's a better way to organize it

components/Navbar.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+const Navbar = () => {
+  return (
+    <div className="bg-neutral-50 p-2 shadow flex justify-between items-center">
+      {/* Logo Section */}
+      <section className="flex items-center space-x-3">
```
anna-murphy: `<section>` is a really confusing element for me... There's an argument that you should just be using a `<div>` here, but I'm not sure! Though, if you want to stick with the `<section>`, MDN says you should have a header for completeness.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section 

Curious to hear your thoughts!

components/Navbar.tsx
```diff
@@ -0,0 +1,278 @@
...
+        </Link>
+      </section>
+
+      {/* Navigation Menu for Desktop */}
+      <NavigationMenu.Root className="hidden lg:block">
+        <NavigationMenu.List className="flex list-none space-x-5">
+          {navigation.map((item) => {
+            return (
+              <NavigationMenu.Item key={item.href}>
+                <NavigationMenu.Trigger className="group inline-flex h-9 items-center justify-center rounded-md bg-background px-4 py-2 text-md font-medium transition-colors hover:bg-accent hover:text-accent-foreground focus:bg-accent focus:text-accent-foreground focus:outline-none disabled:pointer-events-none disabled:opacity-50 data-[active]:bg-accent/50 data-[state=open]:bg-accent/50">
```
anna-murphy: If there's a downside to tailwind, it's that my brain refuses to parse this lol. I'll try and look at it on my machine to give visual feedback, but. that's three lines of classes on my screen 😭.

components/assets/LogoCcv.tsx
```diff
@@ -1,16 +1,16 @@
 import React from "react"
-import { SvgProps } from "@/components/assets/LogoBrown"
+import { LogoProps } from "@/components/assets/LogoBrown"
```
anna-murphy: Hmmm.... This is another organization thing, but I personally don't love importing the props from the `tsx` file. I had to solve this in the riddc app, but what I ended up doing was  file structure like this:

```
components
|
| - Icons
  | - index.ts
  | - types.ts
  | - Brown.svg.tsx
  | - /* whatever other logos */
```
And in `types.ts` I defined the shared properties like `LogoProps` though I think I called it something else. 

I don't want to be a stickler or say that this is the only solution, but I'm just really wary around circular imports (which have, in my experience, been caused by me doing things like this).

So! Not a big problem, but I'd suggest moving it around a little? What do you think?

tailwind.config.js
```diff
@@ -120,14 +120,7 @@ module.exports = {
...
       "2xl": "0 25px 50px -12px rgb(0 0 0 / 0.25)",
-      "book-shadow":
-        "4px 4px 12px 4px rgba(167, 153, 139, 0.35), 12px 12px 4px 0px rgba(167, 153, 139, 0.35);",
-      inner:
-        "-1px -1px 4px 0px rgba(167, 153, 139, 0.35) inset, 1px 1px 4px 0px rgba(167, 153, 139, 0.35) inset",
-      "inner-focus":
-        "-2px -2px 4px 0px rgba(97, 59, 26, 0.35) inset, 2px 2px 4px 0px rgba(97, 59, 26, 0.35) inset",
-      showcase:
-        "4px 4px 12px 4px rgba(167, 153, 139, 0.35), 12px 12px 4px 0px rgba(167, 153, 139, 0.35)",
+      inner: "inset 0 2px 4px 0 rgb(0 0 0 / 0.05)",
```
anna-murphy, line 123: LOVE an inner shadow 👀 

I wonder if we should try and share tailwind configs 🤔

tailwind.config.js
```diff
@@ -78,7 +78,7 @@ module.exports = {
     },
     blur: {
       0: "0",
-      none: "0",
+      none: "",
```
anna-murphy, line 81: Why is this needed? It feels interesting that you wouldn't be able to get this behavior with `blur-0`.

app/layout.tsx
```diff
@@ -18,8 +19,11 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="en">
-      <body className={inter.className}>{children}</body>
-      <Footer />
+      <body className={inter.className}>
+        <Navbar />
+        {children}
```
anna-murphy: I've been thinking about this a lot.... Do we want to put a `<main>` element in the `layout.tsx` or do we want each page to be responsible for it's own `<main>` and accompanying styling?

You have a system that's working for you, but I was feeling like I wanted `<main>` here with styles (depending on whatever kind of layout mode we wanted) and then have each page define only content inside the `<main>`. I haven't been... perfectly consistent about that in the RIDDC project, but. I'm curious to hear if you have thoughts.

components.json
anna-murphy, line 1: What is this file? It looks like it's doing something special but I have no idea!

components/Button.tsx
```diff
@@ -26,7 +26,7 @@ export default function Button({
       break
     case "tertiary":
       buttonStyles =
-        "bg-neutral-50 text-secondary-blue-700 hover:bg-neutral-100"
+        "bg-white text-secondary-blue-700 hover:bg-neutral-100 hover:text-white"
```
anna-murphy, line 29: What's the contrast like on these hover styles?  `neutral-100` feels pretty light when compared with white text... Do you have an example of this on the site? I might be misunderstanding

components/ui/dropdown-menu.tsx
anna-murphy, line 1: Hm. I'll be honest, I'm not sure I'm getting everything about this file lol. You're doing some intricate stuff with `forwardRef` that I think I might be able to review better if I had more context for it -- would you mind walking me through it? I don't doubt that it's fine, I'm more curious at this point!

lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { clsx, type ClassValue } from "clsx"
+import { twMerge } from "tailwind-merge"
```
anna-murphy, line 2: what's `twMerge` doing? and what has your feeling been using `clsx`? I struggle sometimes to know if it's necessary at all 🤔

anna-murphy (approved):
really cool stuff heather! would love to learn about your process during this because you're using some fun techniques. Lots of idle thoughts, but I didn't notice anything earthshattering~

components/Navbar.tsx
hetd54: Given that the footer needs access to this as well, I am in favor of a separate file. The components folder is (eventually) not going to be so flat once everything is written, anyway.

components/Navbar.tsx
hetd54: Truly no thoughts here except that it made it way easier for my brain to picture them as distinct segments when figuring out flexbox. I get confused with the "div div div div" nests. I'll revert to divs here -- they absolutely shouldn't have headings!

components/Navbar.tsx
hetd54: So true!! And using tailwindUI/Shadcn will make this more common

components/ui/dropdown-menu.tsx
hetd54: This is generated from ShadCN: https://ui.shadcn.com/docs/components/dropdown-menu

components.json
hetd54: This is a file added by shadCN, which is shared among all the components created in the UI folder

hetd54: https://ui.shadcn.com/docs/components-json

lib/utils.ts
hetd54: This is added by ShadCn. I'm having trouble finding what it is doing.

s-bessey: twMerge allows you to merge tailwind classes. It looks like this is creating a function that's merging classes (example from above):
```  <DropdownMenuPrimitive.Label
    ref={ref}
    className={cn(
      "px-2 py-1.5 text-sm font-semibold",
      inset && "pl-8",
      className
    )}
    {...props}
  />
```

so the class name is merging a string, `inset && "pl-8"`, and className

components/Navbar.tsx
hetd54: Guh -- this is a copy error from the shadCN trigger. I am in the "solve this with styling" camp and have added "gap-2" to the Trigger!

tailwind.config.js
hetd54: This is an edit by ShadCN -- I had to carefully go through my config and make sure my color editions weren't touched, for example. I'm not sure why they made this decision, but I don't really want to change it until we are finished messing with their components.

components/ui/dropdown-menu.tsx
anna-murphy: Fascinating...

lib/utils.ts
anna-murphy: Hmm I'm not sure. I think `inset` is a boolean as defined in the `dropdown-menu.tsx`, so i think it's`pl-8` if `inset === true` and otherwise `false`.  Which, this is normal behaviour for the `clsx` function. But the thing that i've felt is i'm not certain why this pattern is better than string interpolation. Eg: ```jsx <Ex classname={`${inset ? 'pl-8' : ''}`} />```

So i guess the question is better put as "what does it mean to merge tailwind classes" because I'm not sure about that... does it take a string and search for classes filling the same role and flag them as problematic? 🤷‍♀️  that's the best i could guess

components.json
anna-murphy: interesting! I'm learning so much

lib/utils.ts
s-bessey: Yeah, the docs say that it merges "without causing conflicts" but I don't know how that really differs from string interpolation or how it avoids conflicts...?

This suggests that it determines an order of overrides if two things conflict? https://github.com/dcastil/tailwind-merge/blob/v2.5.5/docs/what-is-it-for.md

PR #22, opened 2024-11-25
Feat: workday api within about page

hetd54:
Screenshot 2024-12-10 at 11 09 18 AM

# Changes

### Folder structure for fetching
- fetching will always occur in the page.tsx server component; the page will then stream the result to child client components
- a queries.js file sits side-by-side with the appropriate page; this file contains all fetch request functions

### Loading States for data fetching
- Within the page.tsx file, the child component that uses the fetched data is surrounded by a <Suspense> tag; there is a fallback for when the data is not done streaming to the child component (a loading spinner)

### SectionHeader component
Gonna use this SO many places so I just made it a component, used in both the About page and Home page


app/about/page.tsx
```diff
@@ -1,3 +1,9 @@
+import { Opportunities } from "@/components/Opportunities"
+
 export default function About() {
-  return <div>About</div>
+  return (
+    <div>
+      <Opportunities />
+    </div>
```
anna-murphy: Do you need the `<div />`?
```suggestion
    <Opportunities />
```

app/page.tsx
```diff
@@ -2,8 +2,8 @@ import EventSection from "@/components/EventSection"
 
 export default function Home() {
   return (
-    <main className="bg-white flex min-h-screen flex-col items-center justify-between p-0 lg:p-24">
+    <div>
       <EventSection />
-    </main>
+    </div>
```
anna-murphy: Maybe remove the `div` here too?

components/LinkCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function LinkCard({
+  className = "",
+  children,
+}: React.PropsWithChildren<CardProps>) {
+  return (
+    <section
+      className={`"h-full p-6 flex flex-col gap-2 rounded-lg drop-shadow-md hover:drop-shadow-lg transition duration-500" ${className ? className : "bg-gray-50 border-white border-2"}`}
+      /* https://design-system.w3.org/components/cards.html#block-link-cards */ data-component="card"
```
anna-murphy, line 14: 👀 
Does this work? I thought you needed some JS in there too to make this work.... BIG if true 🤔

components/Opportunities.tsx
```diff
@@ -0,0 +1,106 @@
...
+      setData(opportunities.json)
+    } catch (err: any) {
+      setError(err.message)
+    } finally {
+      setLoading(false)
+    }
+  }
+  useEffect(() => {
+    handleApiCall()
+  }, [])
```
anna-murphy: Just out of curiosity, why not handle this in one of the server components? That way you only have to transmit the HTML of the page and you don't have to worry about loading state or whatever.

components/Opportunities.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return (
+    <section>
+      <Link href="#opportunities" className="text-neutral-900">
+        <h2 className="text-3xl bg-primary-500 text-white p-4 flex items-center gap-2">
+          <UserPlusIcon className="h-7 w-7" />
+          Opportunities
+        </h2>
+      </Link>
+      {isLoading && <p className="px-2 font-bold">Loading...</p>}
```
anna-murphy: For RIDDC stuff, I recently learned that making a loading spinner is actually really easy! It might be worth looking into if you want keep the loading logic in here. I can link you the code I used for it if you want.

components/Opportunities.tsx
```diff
@@ -0,0 +1,106 @@
...
+      {error && <p className="px-2 font-bold text-red">Error: {error}</p>}
+      {data.jobPostings.length > 0 &&
+        data.jobPostings.map((position) => {
+          return (
```
anna-murphy: This is a very small thing, but when I don't have any Logic happening inside an anonymous function like this, I tend to like to omit the return statement. Now sure if that's your vibe too, but I think the code ends up looking cleaner.

components/Opportunities.tsx
```diff
@@ -0,0 +1,106 @@
...
+                href={`https://brown.wd5.myworkdayjobs.com/en-US/staff-careers-brown${position.externalPath}`}
+              >
+                <LinkCard>
+                  <div className="flex flex-col gap-6 md:items-center md:justify-between md:flex-row">
+                    <div className="flex flex-col gap-2">
+                      <p className="flex items-center text-secondary-blue-700 gap-2 md:gap-4">
+                        <MapPinIcon className="h-4 w-4" /> Providence, RI -
+                        United States
+                      </p>
+                      <p className="text-gray-800">{position.title}</p>
```
anna-murphy, line 43: Wonder if things like this should be `p`'s or `span`'s... MDN didn't come down one way or the other, but I think there might be a reason to use something more generic here, in that this doesn't represent a paragraph in totality? Not sure though!!!


- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span

anna-murphy (changes requested):
Looks pretty good! A few style things, and a general question.

Also, I tried to open the preview link, and I couldn't navigate to the opportunities page... Any thought as to why that was?

anna-murphy (approved):
heck, my mouse slipped at the last minute and i meant to approve.

components/Opportunities.tsx
hetd54: Do you mean pulling this out into its own server component? So "OpportunitiesSection" could be the html and such and Opportunities.tsx could be the server component?

I personally find it frustrating when things get too abstracted and there are 6 files that I need to look into to understand what specifically I need to edit/change; projects get overwhelming as the components folder grows; however, this is running on each page load which is probably overkill? 

Thinking about it, this is worth moving for functionality; I need some help thinking about file structure in the future though!

anna-murphy: Sortof! This is informed by a convention I've come to for myself. When possible, I like to do all my fetching / data interactions at the highest component level I can / is reasonable. For NextJS, it's on the `page.tsx` level. And I think some developers on Next JS and I are of the same opinion about this because all `page.tsx` components are `async` by default!

So the pattern would be that you do your asynchronous fetch action in the page, `await` the result, and then pass the resolved data to the UI component later on. For this example, it might look something like this:

### `app/about/page.tsx`
```jsx
export default async function About() {
  try {
    // Omitting function definition for brevity
    const data: OpportunityProps = await handleApiCall();
    return <Opportunities data={data} />;
  }
  catch (ex) {
    // Handle the case when the fetch fails for whatever reason, display a different page.
    return <p>{ex}</p>;
  }
}
```

### `components/Opportunities.tsx`
```jsx
"use client";
export function Opportunities({ data }: {data:OpportunityProps}) {
  // However you want to display the fetched data, you can omit the fetch effect because you're getting passed validated data.
  return (
    <>
      {data.map( /* ... */)}
    </>
  );
}
```

Does this make sense? I got a little lazy when writing this 😓

app/about/page.tsx
```diff
@@ -1,3 +1,37 @@
...
+    return (
+      <div>
+        <SectionHeader
+          href={"#opportunities"}
+          title={"Opportunities"}
+          icon={<UserPlusIcon className="h-7 w-7" />}
+        />
+        <Suspense fallback={<Spinner />}>
+          <Opportunities data={data} />
+        </Suspense>
```
anna-murphy, line 21: So happy about this pattern!

components/EventSection.tsx
```diff
@@ -24,125 +24,95 @@ function classNames(...classes: (string | boolean | undefined)[]) {
...
-  const [view, setView] = useState("Upcoming")
-  const currentDate = new Date()
-  const today = getStringDate(
-    currentDate.getMonth() + 1,
-    currentDate.getDate(),
-    currentDate.getFullYear()
-  )
+interface EventSectionProps {
+  streamedDataFuture: Promise<any>
+  streamedDataPast: Promise<any>
```
anna-murphy, line 29: Do these have to be `any`?

components/EventSection.tsx
```diff
@@ -24,125 +24,95 @@ function classNames(...classes: (string | boolean | undefined)[]) {
...
-        setDataPast(data)
-      })
-  }, [today])
   return (
-    <article className="container px-2 my-6 space-y-2">
-      <div>
-        <h1 id="events" role="heading" className="font-semibold text-4xl">
-          Events
-        </h1>
+    <div>
```
anna-murphy: can... this be a fragment...?

app/about/queries.js
```diff
@@ -0,0 +1,26 @@
+export async function getWorkdayData() {
+  const myHeaders = new Headers()
+  myHeaders.append("Content-Type", "application/json")
+  myHeaders.append(
+    "Cookie",
+    "PLAY_SESSION=9365d69619d226b22f74256e51894a0476197a1f-instance=vps-prod-tvvtdnej.prod-vps.pr501.cust.pdx.wd; __cf_bm=zYtKNPBpg6veqmLc2qdk_TeHEGdGWIcSRG9tdtkQNbQ-1721323439-1.0.1.1-TApbCV1fQKTtEauZKvZ5XnXrZQJL7dcML6ixA3AycoiBMROa.iAzMq_K.5E7iaZfCj9.WhQQlmYfAxW8wdwZgQ; __cflb=02DiuHJZe28xXz6hQKLf1exjNbMDM5uxekNnt7kFV7LUC; _cfuvid=hqIGrDcFihah5EiFO0c_HEM4gIPwkeWCOxjMxNqjR20-1721323439159-0.0.1.1-604800000; wd-browser-id=fc0f0b1d-4547-488d-b353-99c751c61dae; wday_vps_cookie=1122671626.53810.0000"
```
anna-murphy, line 6: Should this be hidden as a secret? or does it not matter lol

anna-murphy (approved):

app/about/queries.js
hetd54: It wasn't a secret in the old repo, and there are no edit privileges associated with it, so I think it is okay?

PR #24, opened 2024-12-02
Fix routes display main

hetd54:
Firebase was previously re-routing everything to /index.html (my mistake -- copied from another proj); this caused ALL routes in prod to get replaced with the home page content. Redid the firebase hosting config to fix


eldu (approved):
AMAZING! Thanks so much heather !

anna-murphy (approved):
Looks good to me! 


I did notice that it looks like there's some darkmode stuff happening on some of these routes. Obvs, that's not for this PR, but I just wanted to make sure this is on your radar... 
image
(Edit: the nav text is white for some reason)

PR #27, opened 2024-12-05
Status banner

hetd54:
As a reminder, we are using App Hosting now, so no previews. Here's a preview from local -- 

When there are open issues:
Screenshot 2024-12-12 at 9 26 23 AM

When no issues:
Screenshot 2024-12-12 at 9 25 21 AM

We should decide what colors would look nice in the banner -- please give suggestions!


components/header/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+        org: { org },
+        repo: { name },
+        type: "private",
+        sort: "created",
+        direction: "desc",
+      })
+
+      // This currently only gets the last 100 comments on the repo. It should be changed
+      // to get the comments per issue instead, which will be addressed in the next PR
+      const comments = await octokit.request(
```
s-bessey: No need to get the issue comments here!

components/header/Header.tsx
```diff
@@ -0,0 +1,71 @@
+import Banner from "@/components/header/Banner"
+import Navbar from "@/components/header/Navbar"
+import getOpenIssues from "@/components/header/utils"
+
+export default async function Header() {
+  const repositories = await getOpenIssues()
+  // uncomment to test banner with no open issues
```
s-bessey: I'm a little confused about the comment vs uncomment for testing here? Might be good to make that a little clearer

s-bessey (changes requested):
A couple of nitpicks!

hetd54:
Gradient with green and blue colors?
Screenshot 2024-12-05 at 6 12 22 PM
Screenshot 2024-12-05 at 6 22 46 PM

eldu:
@hetd54 @s-bessey 
I think we'll want to streamline the language we're using. It seems like we're using issues/incidents interchangeably. Let's keep it consistent with the status page and the website?

Also,
When all services are operational "View Issues" doesn't seem to make sense since issues could imply something is wrong.
Maybe instead of "View Issues", "View Status"?

s-bessey:
I think the current color blends a bit too much with the rest of the page -- I'd say something a bit darker and without _too_ much contrast between the "top" and "bottom" of the gradient

components/header/Banner.tsx
```diff
@@ -7,6 +7,8 @@ interface BannerProps {
   variant: string
 }
 
+// Try red/orange for the error gradient
```
s-bessey: remove comment?

components/header/Header.tsx
```diff
@@ -1,66 +1,52 @@
 import Banner from "@/components/header/Banner"
 import Navbar from "@/components/header/Navbar"
 import getOpenIssues from "@/components/header/utils"
+import { ExclamationTriangleIcon } from "@heroicons/react/24/outline"
 
+// ONLY HOME PAGE!
+
+// Green/orange dot?? See if that feels like recording
```
s-bessey: Still some notes to self?

s-bessey (approved):
Looking good!

PR #45, opened 2025-04-03
feat: herocard formatting

gtdang:
I updated the hero card: 
* centered component on the page 
* vertically aligned inner cards
* added flex wrapping for the cards 
Screenshot 2025-04-03 at 2 07 40 PM

**Question:**
The linter removed all the semicolons that were once there. Is that standard for this project? I didn't configure anything...

### DIFF
--- a/components/HeroCard.tsx
+++ b/components/HeroCard.tsx
@@ -1,5 +1,5 @@
-import React from "react";
-import { Card, CardContent } from "@/components/ui/card";
+import React from "react"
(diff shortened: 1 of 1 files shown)


ashleysyg:
i also want to know the answer to the semi colons in terms of best practices. i think for now we can just be consistent and remove semi colons across all component files. 

i guess we should also setup a preview link. can i have help with thaat, ellen?

eldu (approved):
This looks great! I tried it out adjusting the size of the screen even.

PR #47, opened 2025-04-08
Chore images

ashleysyg:
replaces anima screenshots with actual images downloaded from figma or shared with group


gtdang (approved):
lgtm

PR #51, opened 2025-04-09
fix: use magic ui for animated hero title

### LINKED ISSUE #31: feat-hero

ashleysyg:


gtdang (approved):
Tested it and it works on my end.

components/magicui/typing-animation.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  useEffect(() => {
+    if (!startOnView) {
+      const startTimeout = setTimeout(() => {
+        setStarted(true);
+      }, delay);
+      return () => clearTimeout(startTimeout);
+    }
+
+    const observer = new IntersectionObserver(
```
eldu, line 41: What is this observer doing?

ashleysyg: in theory, it is making sure the text is in view (10% of viewport) before the animation is triggered to start. i guess that's not necessary here - cause anytime the page is refreshed the hero is going to be in the viewport like 100%... might be worth keeping this component unchanged if we want to use the animated text in some other parts of the site that aren't automatically in view ? it's also the default from magic ui, so keeping it the same will be simpler for other implementations...

PR #53, opened 2025-04-09
Refactor fonts

### LINKED ISSUE #50: fonts
the google font we're using, Source Sans Pro and Source Serif, are all over the place (tailwind config, layout.tsx, globals.css). Probably want to download the fonts and put into assets? and clean up where we define our text variants

ashleysyg:


eldu (approved):
Looks good~

I think for the most part we're using the Source Sans Pro?

Are we using the Serif at all?

ashleysyg:
yea, we use source serif for a few accents in the Impact Banner and Footer (the italics text basically). but source-sans is def used like 99% of the time

PR #58, opened 2025-04-10
refactor: hero

### LINKED ISSUE #54: fix-scoll-to-events
- [x] Make the button in the Hero target scrolling to the Events section instead of the Featured Projects carousel. 
- [x] Consider refactor so unrelated components are not hidden in ClientLayer just for the scrolling action

### LINKED ISSUE #57: refactor-hero
- [x] move hero content out of component, leaving only background image as prop

gtdang:
This refactors the hero component to get rid of the ClientLayout component that we discussed at the 4/9 dev sync. 

Changes:
* `Hero` component now takes children props 
*  Added a `scrollToID` utility function to return a handler for scrolling to document ID
* `HeroHome` component was created for the homepage hero. This was necessary because passing scroll handlers must be client side. 
* Added 'use client' directives to components using react `useState`. This allows import to server-side scripts. 
* Deleted `ClientLayout`


components/Hero.tsx
```diff
@@ -1,53 +1,21 @@
...
-              variant="outline_secondary"
-              className="h-[55px] w-full sm:w-[155px] rounded-[50px] font-semibold text-lg"
-              onClick={onViewEventsClick}
-            >
-              View Events
-            </Button>
-          </div>
-        </div>
-      </div>
+      {children}
```
gtdang: @eldu I made image a prop of this component. Do you think the className should also be exposed?

app/page.tsx
```diff
@@ -20,15 +25,26 @@ export default async function Home() {
...
-          />
-        </Suspense>
+        <div className="relative w-full">
+          <div className="flex flex-col space-y-24">
+            <HeroHome />
+            <HeroCard />
+            <ImpactBanner />
+            <FeaturedCarousel />
+          </div>
+        </div>
```
gtdang: Can I just combine these divs?

app/page.tsx
```diff
@@ -20,15 +25,26 @@ export default async function Home() {
...
+        <div id="events" className="mt-20 w-full">
+          <SectionHeader href={"#events"} title={"Events"} />
```
eldu, line 35: I think now you can delete the `href={"#events"}` ?

components/Hero.tsx
eldu: Nah, the backgroundImage for now should be fine

app/page.tsx
eldu: i think so! try it?

components/HeroHome.tsx
```diff
@@ -0,0 +1,46 @@
+"use client"
```
eldu, line 1: I do like that at least the `"use client"` is in smaller components than the `ClientLayout`. 

A clean up for later could be moving the "use client" to some version of the button component? No need to do that in this PR. This is certainly an improvement! 

https://sentry.io/answers/next-js-13-button-onclick-event-handlers-cannot-be-passed-to-client-component-props/

components/Hero.tsx
```diff
@@ -1,53 +1,21 @@
...
+export const Hero = ({
+  image,
+  children,
+}: {
+  image: string
+  children: ReactNode
+}) => {
   return (
     <div
       className="w-full h-[1200px] bg-cover bg-center relative flex flex-col overflow-hidden m-0 p-0"
```
eldu: I think this `h-[1200px]` is too much?
```suggestion
      className="w-full h-screen bg-cover bg-center relative flex flex-col overflow-hidden m-0 p-0"
```

eldu: `h-screen` means `height: 100vh` which means 100% of the viewport height

eldu: So like 1 screen height

eldu (approved):
approving, i added like a few suggestions

app/page.tsx
gtdang: Ah seems like it's a required prop of SectionHeader. It's ok we'll be getting rid of that SectionHeader component in an upcoming change anyway!

components/Hero.tsx
gtdang: Thanks! Made that change.

PR #61, opened 2025-04-11
refactor: header banner

### LINKED ISSUE #55: Brown Header styling
- [x] brown logo should be in color
- [x] text more prominent (size?, bold?)
- [x] vertical separator missing

gtdang:
Screenshot 2025-04-11 at 2 55 21 PM

* added vertical separator by adding a border to the CCV header 
* moved the "Work with us" button into the same div as the CCV header to get rid of absolute positioning and to follow flex shrinking on smaller screens  

I didn't attempt create a colorized component for the OIT logo. Perhaps better for another PR.


eldu (commented):
Hmm... the vertical line does act weird when I adjust the screen.

I think we can take the header from chatccv? We could keep our button if we wanted

https://ai.ccv.brown.edu/

gtdang:
> Hmm... the vertical line does act weird when I adjust the screen.
> 
> I think we can take the header from chatccv? We could keep our button if we wanted
> 
> https://ai.ccv.brown.edu/

I did the same method as ChatCCV for the separator by adding a left border to the text. I had to change the sizing a bit because our headers are different sizes. You can take a look at the implementation here https://github.com/brown-ccv/chat-ccv/blob/main/components/LandingNav.tsx

ashleysyg (approved):

eldu:
> > Hmm... the vertical line does act weird when I adjust the screen.
> > I think we can take the header from chatccv? We could keep our button if we wanted
> > https://ai.ccv.brown.edu/
> 
> I did the same method as ChatCCV for the separator by adding a left border to the text. I had to change the sizing a bit because our headers are different sizes. You can take a look at the implementation here https://github.com/brown-ccv/chat-ccv/blob/main/components/LandingNav.tsx

@gtdang Ah I'm kind of seeing it now as I slowly adjust screen the chatccv page. Very strange

When the screen gets smaller, let's remove the vertical line and the Center for Computation and Visualization. And as for the logo chatccv has the logo in color so feel free to grab that as well !

eldu:
Screenshot 2025-04-15 at 2 07 10 PM
It's slightly too long when it's on a small screen. It goes past the header height

gtdang:
> > > Hmm... the vertical line does act weird when I adjust the screen.
> > > I think we can take the header from chatccv? We could keep our button if we wanted
> > > https://ai.ccv.brown.edu/
> > 
> > 
> > I did the same method as ChatCCV for the separator by adding a left border to the text. I had to change the sizing a bit because our headers are different sizes. You can take a look at the implementation here https://github.com/brown-ccv/chat-ccv/blob/main/components/LandingNav.tsx
> 
> @gtdang Ah I'm kind of seeing it now as I slowly adjust screen the chatccv page. Very strange
> 
> When the screen gets smaller, let's remove the vertical line and the Center for Computation and Visualization. And as for the logo chatccv has the logo in color so feel free to grab that as well !

Let's discuss this one at today's sync. If we lose the Center for Computation and Visualization that means mobile screens would not have "Center for Computation and Visualization" defined anywhere on our homepage. We could do something like what the [DSI](https://dsi.brown.edu) does where it becomes a sub-banner in place of the nav bar for small screens. The navbar becomes a hamburger and the institute's name "replaces" it.

eldu (approved):

PR #62, opened 2025-04-14
48 feat events section

gtdang:
Screenshot 2025-04-14 at 4 17 03 PM

**Change:**
* Added the card to the left of the events calendar 
* Updated the toggle button order and formatting

To Do (perhaps another PR?): 
- [ ]  Update the text color and day highlight in the calendar 
- [x] Fix weekly-view colors so that grid shows up 
- [ ] Proper flex-wrap for calendar and weekly view


components/EventSection.tsx
```diff
@@ -66,52 +72,78 @@ export function EventSection({
...
-        <div>
-          <CalendarMonth
-            today={today}
-            currentDate={currentDate}
-            events={dataPast.concat(dataFuture)}
-          />
+        <div className="flex-1">
+          {!dataFuture && <p>No event data</p>}
+          {view === "Upcoming" && (
+            <div className="flex flex-wrap justify-between gap-4">
```
eldu: Adding some padding, I know this is slightly different from the figma, but i think it needs some space

```suggestion
            <div className="flex flex-wrap justify-between gap-4 pt-4">
```

components/EventSection.tsx
```diff
@@ -66,52 +72,78 @@ export function EventSection({
...
+                        contact_info={e.contact_info}
+                        description={e.description}
+                      />
+                    </div>
+                  )
+                })}
+            </div>
+          )}
+          {view === "Weekly" && (
+            <div className="h-0 min-h-[768px]">
```
eldu: same here

```suggestion
            <div className="h-0 min-h-[768px] pt-4">
```

components/EventSection.tsx
```diff
@@ -66,52 +72,78 @@ export function EventSection({
...
+              <CalendarWeekly
+                today={today}
+                currentDate={currentDate}
+                events={dataPast.concat(dataFuture)}
+              />
+            </div>
+          )}
+          {view === "Monthly" && (
+            <div>
```
eldu:
```suggestion
            <div className="pt-4">
```

eldu: same here

components/EventSection.tsx
```diff
@@ -66,52 +72,78 @@ export function EventSection({
...
-          <CalendarWeekly
-            today={today}
-            currentDate={currentDate}
-            events={dataPast.concat(dataFuture)}
-          />
+      <div className="flex flex-wrap justify-evenly gap-6">
+        <div className="flex flex-col justify-center items-start min-w-[20%]">
+          <Card className="flex items-start w-full h-full border-none shadow-none">
+            <CardContent className="py-4">
+              <CCVBars />
```
eldu, line 78: In the CCVBars component, I think it should have a margin bottom. It's currently like right up against the events header

`className="mb-4"`

```jsx
import Image from 'next/image'
import CCVBarsImage from '@/assets/CCVBars.png'

export default function CCVBars() {
  return (
    <Image
      className="mb-4"
      src={CCVBarsImage}
      alt="CCVBars"
      width={100}
      height={50}
    />
  )
}
```

eldu (changes requested):
I think overall this looks good! I like that it stacks when I adjust the screen smaller

I added in some padding and margin small edits~

components/EventSection.tsx
ashleysyg: agree. we're def never going to want the ccv bar to go right up against the text under it

ashleysyg:
I have 2 to do items:

- [ ] flex behavior for `Upcoming` view isn't behaving how I might expect/prefer
- [x] monthly view also has problems with grid lines not showing up unless the window size is very narrow

Screenshot 2025-04-15 at 10 46 17 AM
Screenshot 2025-04-15 at 10 52 26 AM


another note: 
the serif font isn't working for "What's next at CCV" but maybe that was an intentional choice to abandon serif?

gtdang:
The Figma had it as a font called Minion Pro.  I didn't see it in the tailwindconfig and just left it out be because I think we'll work on formalizing the styles this week?

ashleysyg:
i think you can add `font-serif` to that text tag, but if that doesn't work then yea it doesn't need to happen for this PR

components/EventSection.tsx
gtdang: Ok. I added this default and the ability to override the default with a custom className. I adjusted margins where CCVBars were used elsewhere.

gtdang:
@ashleysyg @eldu I was able to get the correct alignment of the events card on small screens that we were struggling with yesterday. I also updated the weekly view's background so that the grid shows up. Other calendar colors will need to be adjusted once the global color styles are integrated. Let me know if you're happy with merging!

components/EventSection.tsx
```diff
@@ -5,6 +5,12 @@ import { getStringDate } from "@/components/calendar/utils"
 import CalendarEvent from "@/components/calendar/CalendarEvent"
 import CalendarWeekly from "@/components/calendar/CalendarWeekly"
 import CalendarMonth from "@/components/calendar/CalendarMonth"
+import { Card, CardContent } from "@/components/ui/card"
+import CCVBars from "@/components/assets/CCVBars"
+import { FaCalendarAlt } from "react-icons/fa"
+import { Button } from "./ui/button"
```
ashleysyg: can you make this "@/components/ui/button"?

ashleysyg (approved):
yea looks good to merge - my only note is a minor change to an import to be consistent with other imports - not a functional change. thanks george!!

components/EventSection.tsx
gtdang: done

PR #63, opened 2025-04-15
global styling

### LINKED ISSUE #37: feat-styling-colors

ashleysyg:


tailwind.config.js
```diff
@@ -206,10 +190,28 @@ module.exports = {
         500: "#003c70",
         300: "#527da1",
         50: "#a3b8ca",
+        navbar: "060839",
```
gtdang:
```suggestion
        navbar: "#060839",
```

gtdang (changes requested):

tailwind.config.js
ashleysyg: thansk! done

app/page.tsx
```diff
@@ -2,8 +2,7 @@ import Header from "@/components/header/Header"
 import { HeroHome } from "@/components/HeroHome"
 import { HeroCard } from "@/components/HeroCard"
 import { ImpactBanner } from "@/components/ImpactBanner"
-import SectionHeader from "@/components/SectionHeader"
-import NavbarAnima from "@/components/header/NavbarAnima"
+import { SectionHeader } from "@/components/ui/section-header"
```
gtdang: This import doesn't seem to be used.

components/ui/variants.tsx
```diff
@@ -0,0 +1,123 @@
...
+  {
+    variants: {
+      variant: {
+        // ————————————————————— Primary Filled —————————————————————
+        primary_filled: [
+          "bg-keppel-500 text-white shadow-sm",
+          "hover:bg-keppel-400",
+          "focus-visible:ring-2 focus-visible:ring-keppel-500",
+          "active:bg-keppel-700",
+        ].join(" "),
```
gtdang: I think the style guide has background keppel-700 for the default, 500 for the hover and 600 for the active.

gtdang (changes requested):
Left some comments on the primary button colors. 

I think the button in the EventSection also needs to be updated with the latest variants. 

Otherwise I think it looks good!

PR #71, opened 2025-04-18
Refactor nav

gtdang:
**Changes**
* Moved the navbar out of HeroHome and into `page`. This will allow us to eventually refactor it into `layout` so it will appear on all pages. That will be in an upcoming PR that figures out how to slot the status bar only for the homepage. 
* I made size adjustments to the Hero components. The navbar now sits above the Hero instead of as a transparent overlay so the sizing was off. 
    * The hero has a smaller height. It has different sizes for small vs larger screens. 
    * I updated the HeroCard to overlay above the Hero only on >small screen sizes. It was covering buttons on smaller screen sizes. It now moves below the Hero on smalls screen sizes.


ashleysyg (approved):
my only note is maybe we change NavbarAnima to just Navbar and delete the other Navbar component that we arent using? this is great thanks george!

gtdang:
Additional changes:
* Standardized icons to react-icons package  
* Renamed NavbarAnima -> Navbar and deleted outdated Navbar

Issues:
1. I can't make magnifying glass icon in the search button any larger, the Figma design is much larger than what is currently displayed.

ashleysyg:
> Issues:
> 
> 1. I can't make magnifying glass icon in the search button any larger, the Figma design is much larger than what is currently displayed.

i can fix this when you merge in the feat-styling-colors PR. the new button variants should be able to handle this

PR #78, opened 2025-04-21
feat: add center for computation and visualization to hero

ashleysyg:


gtdang (commented):
Made some changes. Check it out if you think it might need some spacing changes since I got rid of one of the subheadings.

PR #80, opened 2025-04-21
Fix page zoom

### LINKED ISSUE #69: fix-zoom
All the components on the landing page feel too big in Actual View.

ashleysyg:


app/globals.css
```diff
@@ -21,6 +21,19 @@
...
+  main,
+  section {
+    @apply m-0 p-0;
+  }
+}
+
+/* Apply zoom on all screen sizes */
+@media {
+  body {
+    zoom: 0.75;
```
gtdang, line 36: .75 seems a bit too small on my screen. I kind of like 0.95, curious your thoughts.

ashleysyg: i'm open to different zooms for different screen sizes! 0.95 is huge on mine still

ashleysyg: oh interesting... it looks really different on safari and chrome. the text is really small on safari compared to chrome at the same zoom level...

gtdang: Oh... that's not ideal...

gtdang: Comparing to other brown sites like https://engineering.brown.edu/, https://home.watson.brown.edu, https://dsi.brown.edu. Things are just huge... maybe we should just leave it at 1?

ashleysyg: that's true. i'm fine with leaving it full size since it stays in line with other brown sites... i do think some of these commits should be merged into `feat-landing`... like removing fixed sizes in `px` where it's not appropriate (most cases i think...)

gtdang: Yeah I think we can just merge it. Keep the zoom option in there but change it to 1?

PR #83, opened 2025-04-23
Refactor hero

### LINKED ISSUE #52: refactor-status-banner
- [x] status banner becomes a slot / refactoring page vs. layout (nextjs stuff)

This will allow us to refactor and move the heading and navbar out of page and into layout, which will make them appear on all pages of the website.

### LINKED ISSUE #60: refactor "use client" directive to the button level
> I do like that at least the `"use client"` is in smaller components than the `ClientLayout`. 
A clean up for later could be moving the "use client" to some version of the button component? No need to do that in this PR. This is certainly an improvement! https://sentry.io/answers/next-js-13-button-onclick-event-handlers-cannot-be-passed-to-client-component-props/

_Originally posted by @eldu in https://github.com/brown-ccv/ccv-website/pull/58#discussion_r2038069993_

### LINKED ISSUE #66: Header Mobile View
- [x] "Center for Computation and Visualization" added to Hero
- [x] "Center for Computation and Visualization" disappears from Brown Header in mobile view
- [x] check if hero mobile view is correct after merging hero PR

ashleysyg:
- [x] added services landing page because this PR started out as services, but realized that hero should be refactored before new routes are developed
- [x] status banner on landing only
- [x] herohome is becomes generic hero component that goes in app/layout, with image bg and body content dependent on route
- [x] hero home still is client side because hero title animation from magic ui is client side


components/magicui/text-animate.tsx
eldu, line 1: The text animate looks really good!

components/LayoutWithStatusBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+  children,
+  issues,
+}: LayoutWithStatusBannerProps) {
+  const pathname = usePathname();
+  const showStatusBanner = pathname === "/";
+  const repoNames = issues.map((repo) => repo.name).join(", ");
+
+  return (
+    <div>
+      {showStatusBanner && (
```
eldu, line 20: So when there isn't a status banner we have an empty div?

components/LayoutWithStatusBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+import StatusBanner from "@/components/StatusBanner";
+import { Button } from "@/components/ui/button";
+
+interface LayoutWithStatusBannerProps {
+  children: React.ReactNode;
+  issues: any[];
+}
+
+export default function LayoutWithStatusBanner({
+  children,
```
eldu: `children` doesn't seem to be used in this?

app/layout.tsx
```diff
@@ -1,50 +1,62 @@
...
+      <body
+        className={`${inter.className} m-0 p-0 overflow-x-hidden bg-white`}
+      >
+        <LayoutWithStatusBanner issues={issues}>
+          {children}
+        </LayoutWithStatusBanner>
+        <BrownBanner />
+        <Navbar />
+        <div className="flex-grow">{children}</div>
+        <Footer />
```
eldu: `children` looks like it's being used twice here. And then when in `LayoutWithStatusBanner` the children aren't used. I think you can just remove its usage here an have

`<LayoutWithStatusBanner issues={issues}/>`

eldu (changes requested):
The hero looks good! I think the comments i have here are things that aren't actually the hero...

components/LayoutWithStatusBanner.tsx
ashleysyg: changed so when there are no open issues, the status banner becomes green

PR #85, opened 2025-04-29
Fix event section formatting

gtdang:
Updated the event section:
* Added Event Card back to weekly and monthly views
* Updated colors to tailwind config colors
* Increased text size 
* Increased monthly calendar grid gaps 

Issues: 
* ~~I don't love the current stacking on smaller screens. The event card goes between the toggle and calendars. It would probably be better to make everything disappear and revert to "Upcoming" view for smaller screens.~~ [Fixed]

### DIFF
--- a/components/calendar/CalendarWeekly.tsx
+++ b/components/calendar/CalendarWeekly.tsx
@@ -92,14 +92,14 @@ const CalendarWeekly: React.FC<CalendarProps> = ({
           className={`flex items-center justify-center py-3 ${
             isSameMonth(thisDate, activeDate) ? "" : "inactiveDay"
           } ${isSameDay(thisDate, selectedDate) ? "selectedDay" : ""}
-          ${isSameDay(thisDate, currentDate) ? "today bg-sunglow-yellow-50" : "bg-white"}`}
+          ${isSameDay(thisDate, currentDate) ? "today bg-sunglow-50" : "bg-white"}`}
           onClick={() => {
             setSelectedDate(cloneDate)
           }}
         >
-          <span>
+          <span className="text-neutral-900 font-semibold">
             {day}
-            <span className="mx-2 items-center justify-center font-semibold text-gray-900">
+            <span className="mx-2 items-center justify-center">
               {format(thisDate, "d")}
             </span>
           </span>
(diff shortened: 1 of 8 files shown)


components/calendar/CalendarWeekly.tsx
```diff
@@ -133,11 +133,11 @@ const CalendarWeekly: React.FC<CalendarProps> = ({
...
       const calColor =
         event.title === "CCV Office Hours"
-          ? "bg-50 hover:bg-300"
+          ? "bg-keppel-100 hover:bg-keppel-300"
           : event.title ===
               "COBRE CBC Computational Biology Walk-in Office hours"
-            ? "bg-50 hover:bg-300"
-            : "bg-keppel-50 hover:bg-keppel-500"
+            ? "bg-keppel-100 hover:bg-keppel-300"
+            : "bg-keppel-100 hover:bg-keppel-300"
```
gtdang: I'm not really sure why we have this here in the first place. I guess we wanted the option to colorize office hour events differently? This seems like a pain to maintain if the titles change.

ashleysyg: agree. does not seem worth hardcoding this when it's pretty clear they are different office hours based on how diff the titles are

ashleysyg (approved):
looks good! just agree with u on taking out the hardcoded colors for diff office hours

components/ui/tooltip.tsx
```diff
@@ -20,7 +20,7 @@ const TooltipContent = React.forwardRef<
       ref={ref}
       sideOffset={sideOffset}
       className={cn(
-        "z-50 overflow-hidden min-w-48 rounded-md bg-neutral-50 px-3 py-1.5 text-xs animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+        "TooltipContent z-50 overflow-hidden rounded-md bg-neutral-50 px-3 py-1.5 text-lg",
```
eldu: Sorry this is a nit-pick, can the `TooltipContent` class name be in kabob-case like the other class names?

eldu (approved):
approving! I think this looks great! I added in a nitpick comment~ sorry~

components/calendar/CalendarWeekly.tsx
eldu: that's fine! if they were all the same color it's chill unless people want different colors

components/ui/tooltip.tsx
gtdang: Actually didn't even need that! I removed it. It's for if we have custom css styles for defined for tooltip content.

PR #102, opened 2025-05-21
fix: feat-navbar--fix-dropdowns

ashleysyg:


ashleysyg:
When I changed the breakpoint for when the desktop navbar becomes hidden and the hamburger appears to be at a narrow viewport, the desktop navbar dropdown was broken at medium-ish viewports. I ended up implementing a stacked dropdown for when there are 2 or more groups within a navbar item, and that transition happens at the previously broken viewport breakpoint. Ended up fiddling with the padding and borders with lots of tailwind :)

PR #104, opened 2025-05-21
feat: moving content from within components to content folder

ashleysyg:
This PR moves toy data from within About page components into a dedicated `/content` folder. Yaml and Markdown libraries used to read in data from yaml, markdown, and markdown with frontmatter files. `/content/images` is copied into `/public` during build. The People page got some style updates.


package.json
```diff
@@ -23,17 +24,27 @@
     "class-variance-authority": "^0.7.0",
     "date-fns": "^3.6.0",
     "firebase": "^10.12.5",
+    "front-matter": "^4.0.2",
+    "gray-matter": "^4.0.3",
```
eldu: It doesn't like you're using `gray-matter`. Remove it

package.json
```diff
@@ -23,17 +24,27 @@
...
+    "js-yaml": "^4.1.0",
     "lucide-react": "^0.485.0",
+    "markdown-it": "^14.1.0",
     "motion": "^12.9.2",
     "next": "14.2.18",
     "octokit": "^4.0.2",
     "react": "^18",
     "react-dom": "^18",
     "react-icons": "^5.5.0",
+    "remark-frontmatter": "^5.0.0",
```
eldu: same here, doesn't seem to be used

package.json
```diff
@@ -23,17 +24,27 @@
...
+    "remark-parse": "^11.0.0",
```
eldu: same for remark-parse

package.json
```diff
@@ -23,17 +24,27 @@
...
     "tailwind-merge": "^2.5.4",
-    "tailwindcss": "^3.4.10"
+    "tailwindcss": "^3.4.10",
+    "unified": "^11.0.5"
```
eldu: what is this package for?

lib/content-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  } else if (extension === '.yaml' || extension === '.yml') {
+    // dynamically import js-yaml to decrease bundle size and initial load times, 
+    // also can help isolate dependencies and prevent unexpected interactions between parts of code
+    // change if many yaml files and dependency issue isn't a problem
+    const yaml = (await import('js-yaml')).load(fileContent) as Record<string, any>;
+    return { data: yaml, content: '' };
+  }
+}
+
+export async function getAllContent(contentFolder: string) {
```
eldu: Suggesting a rename to something like `getContentFolder` and the parameter be named `dir` or `folder`

eldu: Also is this function copied from the old website? What do you think about a function that took a file path and got that instead of the whole directory? Seems like a lot to pull the whole directory unless it's the images/people folder.

lib/content-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+          };
+        }
+      }
+      return undefined;
+    })
+  ).then(results => results.filter(Boolean) as ({ slug: string; data: Record<string, any>; content: string })[]);
+}
+
+// separate slugify utility
+export function slugify(filename: string): string {
```
eldu: does slugify get used somewhere?

eldu (changes requested):
requesting some changes ~

package.json
ashleysyg: removing it - it supports remark packages

lib/content-utils.ts
ashleysyg: removed!

lib/content-utils.ts
ashleysyg: i changed the about pages to use `readContentFile`. i renamed getAllContent to `readContentFolder`, and it is currently unused but will be useful for the Portfolio pages, where we will be reading in all "software" files or all "workshops" files

lib/content-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export async function readContentFolder(contentFolder: string) {
+  const folderPath = path.join('content', contentFolder);
+  const filenames = await fs.readdir(folderPath);
+
+  return Promise.all(
+    filenames.map(async (filename) => {
+      const filePath = path.join(folderPath, filename);
+      const parsed = await readContentFile(filePath);
```
eldu, line 31: So I think this is actually the fileContent? `parsed` gets used later on as if it's a filename with an extension. Which is correct?

eldu (commented):
This looks really good! I added in one more comment question!!!

lib/content-utils.ts
ashleysyg: parsed is the filecontent... it's used a few lines down to check if it's undefined before checking extensions

eldu (approved):

PR #114, opened 2025-06-05
feat: storage main page

ashleysyg:
This PR adds the Storage main page. 
- It contains the original content from the CCV website. The first section is a lot of text - I preserved even though we will likely change this. 
- All section content is in separate markdown files and they are read in using the `readContentFolder()` utility function.
- Storage Comparison Tool link goes to current website. It will be implemented in another PR.


eldu (approved):
Looks great !

PR #118, opened 2025-06-10
Build: ci build to check lint and typecheck, also update packages

eldu:
run the build step with the necessary tokens to check the lint and typecheck. So we can see if the build fails here rather than checking the app hosting madness

locally you can use `npm run lint` and `npm run typecheck` to do linting and do the typecheck
this run `npn run build` which also checks the linting and typecheck 

also this updates the README


gtdang (commented):
Where is the typecheck or linting being called in the action? Is it part of the build step?

eldu:
> Where is the typecheck or linting being called in the action? Is it part of the build step?

@gtdang It's here https://github.com/brown-ccv/ccv-website/actions/runs/15980179124/job/45072518613?pr=118#step:6:30

`Linting and checking validity of types ...`

If there are errors it will fail the action. 

The action runs `npm run build` which runs `next build`  which checks the linting and typecheck

I tried making the separate jobs for linting and typecheck and it's possible to make a separate job for linting. For typecheck, though it requires the `next-env.d.ts` file which is only created after running dev or build (and then build already does the lint and typecheck)

gtdang:
> > Where is the typecheck or linting being called in the action? Is it part of the build step?
> 
> @gtdang It's here https://github.com/brown-ccv/ccv-website/actions/runs/15980179124/job/45072518613?pr=118#step:6:30
> 
> `Linting and checking validity of types ...`
> 
> If there are errors it will fail the action.
> 
> The action runs `npm run build` which runs `next build` which checks the linting and typecheck
> 
> I tried making the separate jobs for linting and typecheck and it's possible to make a separate job for linting. For typecheck, though it requires the `next-env.d.ts` file which is only created after running dev or build (and then build already does the lint and typecheck)

Ah I see. How is the linting configured? Does it just detect that eslint is in the environment and uses it?

s-bessey:
@gtdang I was just looking this up a couple days ago! Next comes with an eslint plugin by default: https://nextjs.org/docs/pages/api-reference/config/eslint

gtdang (approved):
Looks good. Thanks!

PR #119, opened 2025-06-10
feat: storage tool form

ashleysyg:
This PR creates a storage comparison tool page and the question form that will be used to filter down options/the table. The table will be in a separate PR. I added `storage.yaml` from the OG website, so it should be familiar to Amar and others who already edit the content.

This PR required updates to content utility functions - I was getting typescript errors and it was generally convoluted because of having to handle both markdown and yaml files.  The usage of reading in file or folder content is the same, but the types are more clearly defined. 

All storage tool form types are defined in a separate types files, which is in the `/lib` folder along with the various website utility functions. This was helpful because the storage questions structure in `storage.yaml` is pretty nested.

Should I do this in a separate PR?
- display storage options as simple tags and gray out disqualified ones


eldu (approved):
This looks great!

Thanks for splitting out the storage tool into separate PRs!

For the questions, i think we also wanted some help links. And looks like the table is going to be a separate PR which amazing ! Thanks Ashley !

PR #120, opened 2025-06-11
Feat: storage table

ashleysyg:
This PR adds a table of Storage Services and their Features. 

The table is similar in style to the OG website... but BETTER. Some styling needs work:
- the cell values ideally would all start aligned to the same y distance from the top of the row
- X scroll bars should be visible whenever you need to scroll to see the full table
- the col/row headers need to be sticky
- remove the detail text? the content needs review

The table has the same filtering logic as the OG, but instead of disappearing services/columns, they become disabled/grayed out. 
- the question of data projection replicated vs non replicated combines 2 features... so I'm not sure how that is supposed to work

Mobile friendly view non-existent
- implement cards in another PR


app/services/storage/compare/page.tsx
```diff
@@ -4,28 +4,28 @@ import { Hero } from "@/components/Hero"
 import { TextAnimate } from "@/components/magicui/text-animate"
 import { readContentFile } from "@/lib/content-utils"
 import StorageTool from "@/components/StorageTool"
-import { PageContentData, Question, SelectedAnswers, YAMLQuestionConfig } from '@/lib/storage-types'
+import { PageContentData, Question, SelectedAnswers, YAMLQuestionConfig, YAMLServiceConfig } from '@/lib/storage-types'
 
-const fileName = 'storage.yaml'
-const filePath = path.join(process.cwd(), 'content/services', fileName);
+const fileName = 'storage-tool.yaml'
+const filePath = path.join(process.cwd(), 'content/services/storage', fileName);
```
eldu: Perhaps create a `const` variable for the content folder in your `utils` file. I think it would clean up a bit of these file paths.

Something like 
```ts
const contentFolder = path.join(process.cwd(), "content")
```

content/services/storage/brown-storage-options.md
```diff
@@ -2,7 +2,7 @@
 title: Brown Storage Options
 links: 
   - text: "Compare Storage Options"
-    href: "https://ccv.brown.edu/storage/"
+    href: "storage/compare"
```
eldu, line 5: ❤️

content/services/storage/storage-tool.yaml
```diff
@@ -228,7 +228,7 @@ services:
       - name: storage_warmth
         class: cold
 
-  - name: LabArchives
+  - name: Lab_Archives
```
eldu: Are we removing work separators in this file? Looks like the `Campus File Storage Non-replicated` is now spaces

```suggestion
  - name: Lab Archives
```

lib/utils.ts
```diff
@@ -40,3 +40,15 @@ export const scrollToID = (targetId: string) => {
   const targetElement = document.getElementById(targetId)
   return targetElement?.scrollIntoView({ behavior: "smooth" })
 }
+
+export const humanize = (str: string | null | undefined): string => {
```
eldu, line 62: Is this from the old website? Anyway, can you add in a docstring for what this function is doing? For the utils functions docstrings are super useful!

components/ui/table.tsx
```diff
@@ -0,0 +1,300 @@
...
+        continue; // Skip this filter if selected answer option is missing
+      }
+
+      const allowedCategoryClasses = selectedYAMLAnswerOption.category_classes;
+
+      const serviceFeature = service.features.find(f => f.name === questionId);
+
+      // Determine if the selected answer implies a *strict requirement* for the feature.
+      const nonStrictAnswers = ['no risk', 'no', 'any', 'not sure', 'less than 1 tb']; 
+      const isStrictFilter = !nonStrictAnswers.includes(String(selectedAnswerValue).toLowerCase());
```
eldu, line 96: I'm not a big fan of implying from a answer. Is there a way to be more explicit in one of the question configs?

components/ui/table.tsx
```diff
@@ -0,0 +1,300 @@
...
+        console.log(`-> DISABLED: Feature '${questionId}' with class '${serviceFeatureClassNormalized}' does NOT match allowed classes.`);
+        console.groupEnd();
+        return true;
+      }
+      
+      console.log(`-> PASSED: Feature '${questionId}' with class '${serviceFeatureClassNormalized}' matches allowed classes.`);
+      console.groupEnd();
+    }
+  }
+  console.log(`[Filtering Debug] Service: ${service.name} PASSED ALL FILTERS. Showing.`);
```
eldu, line 136: there are a TON of console logs. It doesn't have to be in this PR, but we'll want to remove these eventually

components/ui/table.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+          const featureClassLower = String(feature.class).toLowerCase();
+          const IconComponent = FeatureIcons[feature.name.toLowerCase()];
+          const valueColor = ClassColors[featureClassLower] || ClassColors['default'];
+
+          return (
+            <div className={cn("flex flex-col items-center justify-center p-2 min-h-[80px]", cellContentClasses)}>
+              <div className={cn("flex gap-2 font-semibold text-xl")}>
+                {/* Conditionally render IconComponent to avoid errors if it's undefined */}
+                {IconComponent && <IconComponent className={cn("text-2xl", valueColor)} />}
```
eldu: This is a style choice really. I'm more of a fan of the following conditional render. Sometimes the `&&` could run you into trouble if the left side is a number. For example if you had something like `messageCount && <p>New messages</p>` and `messageCount` was `0` it would render `0`. 

```suggestion
                {IconComponent ? <IconComponent className={cn("text-2xl", valueColor)} /> : null}
```

See https://react.dev/learn/conditional-rendering for more!

components/ui/table.tsx
```diff
@@ -0,0 +1,300 @@
...
+                <th
+                  key={header.id}
+                  colSpan={header.colSpan}
+                  className={cn(
+                    "px-2 py-3 text-left text-md font-medium text-neutral-500 uppercase tracking-wider",
+                    header.column.getCanSort() ? 'cursor-pointer select-none' : ''
+                  )}
+                  style={{ width: header.getSize() }}
+                >
+                  {header.isPlaceholder
```
eldu: What is the placeholder?

eldu (changes requested):
Added some comments and some questions

Also from the GSDC meeting
- shorten the question as much as possible with relevant links.

I think a separate PR could be for the following
- In the main storage page, make the storage options bullet point list into links that link to the sections underneath the compare table.

And other explorations expressed in the gsdc meeting
- What if we flipped the rows and columns?
- What if we filtered more directly in the table? So like in the column filter or in a row filter. 
- Keep mobile development in mind
- What if the questions were ordered the same as the table rows?

I think for the next chunk, figure out a table to card situation for mobile. And we can revisit the table again another time (don't get too deep into a rabbit hole with the table changes)

content/services/storage/storage-tool.yaml
ashleysyg: i copied and pasted the OG storage yaml to make sure it is consistent with what the storage team has worked with before. Removes some of these changes that I had made while debugging

lib/utils.ts
ashleysyg: yes!

app/services/storage/compare/page.tsx
ashleysyg: requires a rewrite of the readContentFolder function that I didn't want to deal with right now. i just made the filePath the full string instead of breaking it up into 3 lines for now

components/ui/table.tsx
ashleysyg: it is optional for when you have multi level headers. we don't so i removed this

components/ui/table.tsx
ashleysyg: will do this in another PR! i made an issue

PR #126, opened 2025-06-18
Feat add cobre header text

gtdang:
* Added "COBRE-Center for computational Biology of Human Disease" to the header
* Added the header text to the navbar on smaller screens because the header text in the banner hides on small screens.


gtdang:
Wrong repo!

PR #128, opened 2025-06-18
Fix  storage types

### LINKED ISSUE #121: storage exploration
- [ ] Filter more directly in the table - like in the column filter or in a row filter?
- [x] consolidate storage yamls into one 
- [x] link to the storage service sections by clicking on the storage table header
- [ ] always show scroll bars, make header stcky

ashleysyg:
This PR fixes the bug in the main storage page (`services/storage`). 
- The issue was it was reading in the entire storage content folder before `storage-tool.yaml` file was added. This caused downstream bugs when looping through the content files because the keys/values are different from the main page content and the storage tool content. 

```
- Changes:
- combine all storage content into `storage-tool.yaml` because it will be easier to update/maintain.
- clean up and fix interfaces and props in all storage components
- put color/icon mapping in the storage types file, and import into storage tool components (needed to add a whitelist to the tailwind config for custom colors - explanation for this is in the tailwind config)
```


components/StorageServiceCard.tsx
```diff
@@ -1,146 +1,89 @@
...
-const ServiceCard: React.FC<ServiceCardProps> = ({ service, yamlQuestionsConfig, selectedAnswers, isDisabled }) => {
+const StorageServiceCard: React.FC<ServiceCardProps> = ({ service, questionsConfig, selectedAnswers, isDisabled }) => {
 
   // Helper to determine if a service feature's class matches the current selected answer
-  const matchesSelectedAnswer = (featureName: string, featureClass: string | boolean | number): boolean => {
-    const selectedValue = selectedAnswers[featureName];
-    if (!selectedValue) {
-      return true; // No filter for this feature, so it matches
-    }
+  // const matchesSelectedAnswer = (featureName: string, featureClass: string | boolean | number): boolean => {
```
eldu: There is a lot of commented out code here?

eldu: Does it need to be here or can we remove it?

components/ui/form.tsx
```diff
@@ -1,35 +1,44 @@
...
+            <Markdown
+              rehypePlugins={[rehypeRaw]}
+              remarkPlugins={[remarkGfm]}
+            >
+              {question.information}
+            </Markdown>
+            )
+          }
+
+          {/* {question.information && question.information.length > 0 && (
```
eldu: same question about commented out code here

eldu (changes requested):
added some questions about the commented out code

components/StorageServiceCard.tsx
ashleysyg: removed!

components/ui/form.tsx
ashleysyg: also removed

ashleysyg:
I could make a folder `/components/storage` similar to the calendar components and move the storage components into it?

eldu:
> I could make a folder `/components/storage` similar to the calendar components and move the storage components into it?

@ashley i think that would be nice !

eldu (approved):
approving

PR #129, opened 2025-06-19
chore: make people images square

### LINKED ISSUE #112: Fix people photos
Currently we're squishing them to fit in the circle, but it looks kind of bad. It would be great if people uploaded a square photo! 

To help encourage square photos some ideas are -- 
* github action to make sure that the photos are square? -- force a square photo
* maybe there is some threshold of photos that is okay
* maybe always keep the correct aspect ratio and get the circle in the center of the photo

ashleysyg:
resize all people images that were not square manually and update build-utils to check if public/images folder exists and if so, delete before copying


lib/build-utils.ts
eldu, line 1: Did you do this because it wasn't overwriting?

eldu (approved):
Thanks for fixing up the images! I added in a question

lib/build-utils.ts
ashleysyg: yes, it wasn't overwriting and also was not removing images if they were deleted since the last build. i wanted it ideally to work like rsync - just changing files/folders within content/images that are different and also adding/removing files/folders. but, this was the simplest way to accomplish that and I figured it's okay because there aren't actually that many images so copying the whole thing would be okay?

PR #130, opened 2025-06-19
fix: expand/make generic the interfaces for storage and about pages (so far)

### LINKED ISSUE #124: feat: refactor content-utils and pageContentData for non-storage content
contact us page, for example, uses readContentFile(), which takes pageContentData. pageContentData currently has types defined for storage tool data.

ashleysyg:
This PR makes a PageContentData for each subroute (storage, about, etc). It also updates the content-utils to assume that the content folder will never have non yaml/yml/md files

### DIFF
--- a/app/about/contact/page.tsx
+++ b/app/about/contact/page.tsx
@@ -8,21 +8,10 @@ import { cn } from "@/lib/utils";
 import { cardVariants } from "@/components/ui/variants";
 import { readContentFile } from "@/lib/content-utils"
 import Icon from "@/components/ui/render-icon";
+import { ContactUsTypes, OfficeHoursTypes, PageContentData } from "@/lib/about-types";
 
-interface ContactTypes {
-  title: string;
-  icon?: string;
-  description: string;
-  buttonLinks?: { text: string; href: string }[];
-}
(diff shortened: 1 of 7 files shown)


app/about/contact/page.tsx
```diff
@@ -57,7 +46,7 @@ export default async function ContactUs() {
 
         <div className="content-wrapper flex justify-center px-40">
           <div className="flex flex-wrap justify-center gap-y-6 gap-x-6 xs:w-1/2">
-            {pageContent?.data?.contactUs?.map((card: ContactTypes) => (
+            {pageContent?.contactUs?.map((card: ContactUsTypes) => (
```
eldu, line 49: In all of these page contents, there is a lot of `?` for whether its undefined. But we know that it isn't undefined. 

I think there is a better way to define the types such that there isn't as much undefined nature in the way you access the data. let me know if you want to take a look at that together at some point

eldu (approved):
I'm approving to keep this going. I did add in a comment about the typing

app/about/contact/page.tsx
ashleysyg: yes! i'll create an issue and would love some help with that.

PR #131, opened 2025-06-25
Feat(storage):  services header routing

ashleysyg:
This PR turns the Services in the Storage Table header into next links that re-route to service info in main storage page. So clicking on Google Drive takes you to the Google Drive section of `/services/storage` and so on.

Also found a bug in some of the Feature icons and in the text for `Globus` and `Campus File Storage/Department File Services`. I changed the text to say "-" rather than "Undefined", but maybe we want to fill out the cell values in `/content/services/storage/storage-tool.yaml`?

### DIFF
--- a/app/services/storage/page.tsx
+++ b/app/services/storage/page.tsx
@@ -61,7 +61,7 @@ export default async function Storage() {
           const sectionBgColor = isEven ? 'bg-white' : 'bg-neutral-50';
 
           return(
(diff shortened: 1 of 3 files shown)


eldu (commented):
I think this PR has changes from another PR? What branch is this supposed to be going into?

ashleysyg:
> I think this PR has changes from another PR? What branch is this supposed to be going into?

Oh yes you are right. It goes into `fix--storage-types`

PR #135, opened 2025-06-26
fix(style): horizontal scroll buttons in table header

ashleysyg:
This PR adds a horizontal scroll bar that is always present for the table because it is not always clear that the table has more services when the column edge is aligned with teh table edge. It also fixes a minor style bug where the top of the table and top of the form should be aligned when they are side by side (at size XL and above).


components/storage/StorageTableScrollButton.tsx
eldu, line 1: Style question -- Is there anyway this could be like an circle button with an icon inside?

eldu (approved):
approving! i added in a style comment

components/storage/StorageTableScrollButton.tsx
ashleysyg: Yes! I discovered issues with non font awesome icons - so i changed any non Fa icons to Fa. I made changes to the button variant sizing to reflect this. I could not solve the issue with some icons in the first column of the table so I gave up. Made an issue.

PR #137, opened 2025-06-26
DELETE-Feat: storage table order same as form question order

ashleysyg:
This PR sorts the features in the table the same as the form questions. The features that are not in the form questions are sorted alphabetically at the end.

### DIFF
--- a/components/storage/StorageTable.tsx
+++ b/components/storage/StorageTable.tsx
@@ -79,25 +79,52 @@ const Table: React.FC<TableProps> = ({ services, selectedAnswers, questions }) =
   const columnHelper = createColumnHelper<TableRow>();
   const tableContainerRef = useRef<HTMLDivElement>(null);
 
-const tableData: TableRow[] = useMemo(() => {
-  const uniqueFeatureNames = new Set<string>();
-  services.forEach(service => {
+  // --- FEATURE SORTING LOGIC ---
+  const sortedFeatureNames = useMemo(() => {
+    const featureNamesFromForm = questions.map(q => q.affected_feature);
+    const uniqueServiceFeatureNames = new Set<string>();
+    services.forEach(service => {
       service.features?.forEach(feature => {
-          uniqueFeatureNames.add(feature.name);
+        uniqueServiceFeatureNames.add(feature.name);
       });
-  });
+    });
 
-  // TODO: sort the table rows by questions first, then other features alphabetically
-  const orderedFeatureNames = Array.from(uniqueFeatureNames).sort();
+    const featuresNotInForm: string[] = [];
+    const featuresInFormOrder: string[] = [];
 
-  return orderedFeatureNames.map(featureName => {
+    // Separate features: those in the form, and those not
+    Array.from(uniqueServiceFeatureNames).forEach(featureName => {
+      if (featureNamesFromForm.includes(featureName)) {
+        featuresInFormOrder.push(featureName);
+      } else {
+        featuresNotInForm.push(featureName);
+      }
+    });
+
+    // Sort features that are in the form based on their order in questions
+    featuresInFormOrder.sort((a, b) => {
+      const indexA = featureNamesFromForm.indexOf(a);
+      const indexB = featureNamesFromForm.indexOf(b);
+      return indexA - indexB;
+    });
+
+    // Sort features not in the form alphabetically
+    featuresNotInForm.sort((a, b) => a.localeCompare(b));
+
+    // Combine them: form features first, then others alphabetically
+    return [...featuresInFormOrder, ...featuresNotInForm];
+  }, [services, questions]);
+
+  const tableData: TableRow[] = useMemo(() => {
+    // Use the newly sortedFeatureNames
+    return sortedFeatureNames.map(featureName => {
       const row: TableRow = { featureName: featureName };
       services.forEach(service => {
-          row[service.name] = service.features?.find(f => f.name === featureName);
+        row[service.name] = service.features?.find(f => f.name === featureName);
       });
       return row;
-  });
-}, [services]);
+    });
+  }, [services, sortedFeatureNames]);
 
   // 2. define columns for TanStack Table (memoized for performance)
   const columns = useMemo<ColumnDef<TableRow, any>[]>(() => {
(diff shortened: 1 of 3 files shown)


components/storage/StorageTable.tsx
```diff
@@ -79,25 +79,52 @@ const Table: React.FC<TableProps> = ({ services, selectedAnswers, questions }) =
   const columnHelper = createColumnHelper<TableRow>();
   const tableContainerRef = useRef<HTMLDivElement>(null);
 
-const tableData: TableRow[] = useMemo(() => {
-  const uniqueFeatureNames = new Set<string>();
-  services.forEach(service => {
+  // --- FEATURE SORTING LOGIC ---
+  const sortedFeatureNames = useMemo(() => {
```
eldu, line 88: I think i'd like to walk through this function with you

Why do you need the `useMemo`?

components/storage/StorageTable.tsx
```diff
@@ -79,25 +79,52 @@ const Table: React.FC<TableProps> = ({ services, selectedAnswers, questions }) =
...
+    const featureNamesFromForm = questions.map(q => q.affected_feature);
+    const uniqueServiceFeatureNames = new Set<string>();
+    services.forEach(service => {
       service.features?.forEach(feature => {
-          uniqueFeatureNames.add(feature.name);
+        uniqueServiceFeatureNames.add(feature.name);
       });
-  });
+    });
```
eldu: Is the list of features (i'm assuming you mean the rows) not the same for all of the services? We could just define that all of the services need to list something for the feature. And then you wouldn't need to create a set

components/storage/StorageTable.tsx
```diff
@@ -79,25 +79,52 @@ const Table: React.FC<TableProps> = ({ services, selectedAnswers, questions }) =
...
-  // TODO: sort the table rows by questions first, then other features alphabetically
-  const orderedFeatureNames = Array.from(uniqueFeatureNames).sort();
+    // Separate features: those in the form, and those not
+    Array.from(uniqueServiceFeatureNames).forEach(featureName => {
+      if (featureNamesFromForm.includes(featureName)) {
+        featuresInFormOrder.push(featureName);
+      } else {
+        featuresNotInForm.push(featureName);
+      }
+    });
```
eldu: In this part, you could use this `filter` function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

You might only need the to filter for the ones that are not in the form?

components/storage/StorageTable.tsx
```diff
@@ -79,25 +79,52 @@ const Table: React.FC<TableProps> = ({ services, selectedAnswers, questions }) =
...
 
-  return orderedFeatureNames.map(featureName => {
+    // Sort features that are in the form based on their order in questions
+    featuresInFormOrder.sort((a, b) => {
+      const indexA = featureNamesFromForm.indexOf(a);
+      const indexB = featureNamesFromForm.indexOf(b);
+      return indexA - indexB;
+    });
```
eldu: Does the form already have the order though?

eldu (changes requested):
I'd like to walk through this function with you. it's really funky

components/storage/StorageTable.tsx
ashleysyg: it is the same for the services that have features, but Globus does not have any features. i wanted to leave Globus in `storage-tool.yaml` as a service, rather than putting it in a separate markdown file. Globus should not be in the table, because it is not a service per say. The way I remove it from the table is removing it from the table if it doesnt have any features. maybe it makes more sense to have it in it's own markdown file?

ashleysyg: they are the same for all services, except for globus. i know we clarified in the last meeting that globus is not a service... but i wanted to keep it in the same single storage-tool content file, rather than making a one off markdown for globus. i am open to doing that though... then yea this isnt necessary.

components/storage/StorageTable.tsx
ashleysyg: yes!

components/storage/StorageTable.tsx
ashleysyg: i clarified the comments for this part because i think that is (part of) what is making it confusing... it's sorting the features from the `Services` part of the yaml. They are not in the same order as the questions. The questions start with `security`, `DOI_provided`, etc. Here is the order of the feautres for Stronghold:

```
- name: stronghold
    description: |
      [Stronghold](https://it.brown.edu/services/stronghold-research-environment-data-compliance) is a secure computing and storage environment that enables Brown researchers to analyze sensitive data while complying with regulatory or contractual requirements.

      - Best for: Storing data with data usage agreements, FISMA, etc.
      - Rate: $100/TB/Year when storing above free/grant allocations
    links:
      - text: Request Service
        target: https://www.brown.edu/cis/forms/Stronghold/shold.php
    features:
      - name: relative_speed
        value: faster
      - name: collaborative_edits
        value: False
      - name: shareable_link
        value: False
      - name: DOI_provided
        value: False
      - name: data_protection_snapshots
        value: True
        notes:
          - up to 6 weeks
      - name: data_protection_replication
        value: True
      - name: canvas_integration
        value: False
      - name: brown_network_required
        value: True
        notes:
          - except with Globus
      - name: access_from_oscar
        value: False
      - name: security
        value: 3
        notes:
          - Appropriate for data with strong data compliance requirements PII, HIPAA, etc
      - name: storage
        value: 1 TB +
        notes:
          - 1 TB per PI
          - 10 TB per grant
          - Paid after that @ $100/TB/Yr. (Summed across all paid storage types)
      - name: max_file_size
        value: 4 TB
        notes: 
          - 1 TB for best performance
      - name: storage_warmth
        value: hot
```

components/storage/StorageTable.tsx
ashleysyg: well, i think given that the table takes so long to load it maybe isn't really helping at all. but i thought that it is supposed to help with performance. so at each time the page re-renders, `useMemo()` only runs that code whenever the dependencies have changed. So in this case, all of this sorting code only runs if `sortedFeatureNames` has changed

ashleysyg:
I think I can delete this PR because [this one](https://github.com/brown-ccv/ccv-website/pull/143) does the re-ordering much better.

eldu:
Yeah, let's delete this one !

PR #138, opened 2025-06-26
feat: makes the row and column headers sticky

ashleysyg:
This PR makes the storage table column header and row header sticky when you scroll in either x or y direction.


eldu (changes requested):
There are some small styling issues. Could you take a look?

Screenshot 2025-06-30 at 3 01 11 PM
Screenshot 2025-06-30 at 3 00 22 PM
Screenshot 2025-06-30 at 2 59 46 PM

ashleysyg:
fixed! i removed py to get rid of the scrolling excess between the header and the div containing the scroll buttons above it. made the lines sticky when scrolling!

eldu (approved):
amazing!!! thanks so much ashley !

PR #139, opened 2025-06-26
DELETE-Fix  remove storage table columns

ashleysyg:
This PR updates the storage-tool.yaml, removing Campus Department Files and moving Globus to the top of Services because it is not really a "service", but we want the text to be easy to find for updating. It also adds logic to only include Services with the Features key (it should have at least one feature). If there are some missing features for that service, the table displays a dash.


eldu (changes requested):
So this PR is going into the `feat-style--storage-table-scroll` branch. However it looks like it has stuff from [feat-style--storage-table-order](https://github.com/brown-ccv/ccv-website/tree/feat-style--storage-table-order). 

This PR should only bee the yaml file changes

PR #142, opened 2025-07-02
Feat style storage cards

ashleysyg:
This PR makes style changes to the storage cards and refactors the badge component. It was originally written for the featured carousel, so the variants and color mappings had to be updated. There is now also an auto-styling prop for the badge and a helper function to decide if the badge text should be black or white (roughly following WCAG accessibility guidelines). There is an added badge usage readme and examples.

Main style changes are:
- turning the feature values into badges, so they are easier to read vs. colored text
- changing the size of the features to have a more consistent look with the table
- putting the feature notes in their own row beneath the feature and values, and making them bulleted lists rather than a string concatenating the items in the notes array


components/ui/badge-examples.tsx
eldu, line 1: I don't think we need this file? The mini explanation on top of the Badge component seems sufficient

```
/**
 * Enhanced Badge component with automatic color determination
 * 
 * @example
 * // Traditional usage with predefined colors
 * <Badge color="keppel">Success</Badge>
 * 
 * @example
 * // Automatic color based on feature value
 * <Badge value="high" autoColor>High Priority</Badge>
 * <Badge value={true} autoColor>Available</Badge>
 * <Badge value="fast" autoColor>Fast Speed</Badge>
 */
 ```

File: components/ui/README-badge-system.md
Comment (eldu) [inline components/ui/README-badge-system.md:1]: Wow! Thanks for the documentation. Could more of this information be put into the `badge.tsx` file? To keep in the information closer to the code?

File: components/ui/README-badge-system.md [hunk 1]
```diff
@@ -0,0 +1,179 @@
...
+
+Maps feature values to background colors:
+
+```tsx
+import { getBadgeBackgroundColor } from '@/lib/utils';
+
+const bgColor = getBadgeBackgroundColor('high');
+// Returns: 'bg-red-university'
+```
+
```
eldu: This documentation could be on top of those utility functions

components/ui/variants.tsx
eldu, line 1: We're going to have a bunch of variants for different components. Are they all going to live in this file? 

What do you think about if the various were in the same file as the original components? Does that work? Thinking about organizing~~

eldu (changes requested):
These badges LOOK sooo good! Great job! 

I have some refactoring organization suggestions. Nothing really more than that. Mainly trying to keep the documentation closer to the code to help the likely-hood of upkeeping documentation 😅

components/ui/variants.tsx
ashleysyg: Moving the storage variants, for example, to `StorageTool.tsx`?
And leaving the ones that are re-used in different components in `variants.tsx`?
I feel good about that

components/ui/README-badge-system.md
ashleysyg: yes! moved the most critical documentation to `badge.tsx` and deleted the extra README and examples files

components/ui/badge.tsx
```diff
@@ -23,6 +48,26 @@ type BadgeProps = React.HTMLAttributes<HTMLDivElement> &
...
+ *   import { getTextColorForBackground } from '@/lib/utils';
+ *   const textColor = getTextColorForBackground('bg-keppel-600');
+ *   // Returns: 'text-white'
+ *
+ * getBadgeBackgroundColor(value)
+ *   Maps feature values to background colors:
+ *   import { getBadgeBackgroundColor } from '@/lib/utils';
+ *   const bgColor = getBadgeBackgroundColor('high');
+ *   // Returns: 'bg-red-university'
  */
```
eldu: I think this can be deleted now since it's on top of the util functions in the `storage/utils.ts`

eldu (approved):
approving, i added in one more comment to remove some text since it's already in place elsewhere

PR #143, opened 2025-07-02
feat: reorder storage tool features (both desktop/table and mobile/card views)

ashleysyg:
This PR changes the order of features so they match the questions in the storage form. It alphabetizes features not in the form under the matching features. It uses improved/simplified logic compared to [the storage table reordering PR](https://github.com/brown-ccv/ccv-website/pull/137). Both the cards and table views are sorted properly now. 

This PR also removes services if they do not have features in `storage-tool.yaml` (this is relevant for Globus only)


components/storage/StorageServiceCard.tsx
```diff
@@ -24,6 +24,32 @@ const StorageServiceCard: React.FC<ServiceCardProps> = ({ service, isDisabled })
     return String(feature.value);
   };
 
+  // Helper to sort features according to question order, then alphabetically
+  const sortFeatures = (features: ServiceFeature[]): ServiceFeature[] => {
```
eldu: Oohhh this one is sooo much easier to understand. Also I think you're defining this twice in this PR. Can there just be one definition that you use to import?

eldu (changes requested):
Small change requested! I think it also loads faster!!!

components/storage/StorageServiceCard.tsx
ashleysyg: yes! it loads so much faster :)

components/storage/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    service.features?.forEach(feature => {
+      uniqueFeatureNames.add(feature.name);
+    });
+  });
+  return Array.from(uniqueFeatureNames);
+}
+
+/**
+ * Sorts features or feature names according to the order in questionsConfig, then alphabetically.
+ * Accepts an array of features ({ name: string }) or feature names (string[]), and a questionsConfig array.
```
eldu, line 20: Interesting... i guess i didn't realize that it was taking in different kinds of input. It would be nice if it only had to handle one type 😕 

this is kind of wonky, but it's also only used on the storage page... so i'ma let it go. lol

eldu (approved):
approving, i added in a comment suggestion?

components/storage/utils.ts
ashleysyg: tysm

PR #144, opened 2025-07-02
chore: storage content updates

ashleysyg:
This PR updates the storage content from Amar's Google Doc and puts hero links into 2 rows for the two new link categories. This PR replaces the content updates in [this PR](https://github.com/brown-ccv/ccv-website/pull/139) - which are now out of date. I handle the conditional rendering of services with and without features in a different PR.


eldu (approved):
Approving for the content changes

I don't superly love the spacing of the rows of buttons... but i'm okay lol. Maybe if the spacing between the button rows is smaller? idk. 

I also think the color of the buttons should be more consistent across the whole website. Were we defining the button colors based on whether they were internal/external links? Or pure style looks based on the situation?

ashleysyg:
Currently defining button colors based on pure style. I'm open to suggestions... I also hate the spacing of the button rows...

PR #154, opened 2025-07-02
chore: gets navbar ready for launch

### LINKED ISSUE #149: chore(navbar): clean up for launch
- [ ] Remove Blog
- [ ] "Coming Soon" for Portfolio
- [ ] Make Publications link open in new tab
- [ ] comment out Search

ashleysyg:
This PR comments out blog and search, adds "coming soon" to portfolio, adds icons to portfolio, and makes link to Publications open in another tab.


components/Navbar.tsx
```diff
@@ -367,6 +381,7 @@ const NavigationSectionContent: React.FC<{
                 <li key={route.href} className="hover:bg-neutral-50 focus:bg-neutral-100 p-1 sm:p-1.5 md:p-2">
                   <NavigationMenu.Link
                     href={route.href}
+                    {...(route.href === "https://publications.ccv.brown.edu" ? { target: "_blank", rel: "noopener noreferrer" } : {})}
```
eldu, line 395: Rather than this checking if this is the publications link, making a prop somewhere that's like `external={true}` that would then add in the `{ target: "_blank", rel: "noopener noreferrer"}` fields would be more generalized!

components/Navbar.tsx
```diff
@@ -321,23 +329,29 @@ export const Navbar: React.FC = () => {
                     {section.groups.map(group => (
                       <div className="py-2" key={group.name}>
                         {group.routes.map(route => (
-                          <Link key={route.href} href={route.href} className="block text-white text-xl py-6 px-6 mr-6 hover:bg-sunglow-400 hover:text-black active:bg-sunglow-200">
-                            {route.name}
-                          </Link>
+                          route.href === "https://publications.ccv.brown.edu" ? (
```
eldu: same here... i wouldn't just be checking if it matches the publications link (string not even a variable!!!) 

In the route config have like an internal/external key for true/false???

eldu (changes requested):
requesting a refactor

ashleysyg:
I made a separate component for links that takes external as a prop and replaced all <a> and <Link> tags with it

components/ui/external-link.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type ExternalLinkProps = React.PropsWithChildren<{
+  external?: boolean;
+  className?: string;
+}> & Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'href'> & Omit<LinkProps, 'href'> & {
+  href: string;
+};
+
+export const ExternalLink: React.FC<ExternalLinkProps> = ({
+  external = false,
```
eldu: I'd suggest renaming this component to `Link` or something like that. If it is named `ExternalLink` I would believe that this prop `external` should default to `true`

eldu (changes requested):
One more request change ! Thanks Ashley for the refactor !

components/ui/external-link.tsx
ashleysyg: i went with the latter option because Link is already a Next name and most of our links are external anyway

eldu (approved):

PR #167, opened 2025-07-09
refactor: About Us

ashleysyg:
This PR moves hardcoded text into yaml

### DIFF
--- a/app/about/contact/page.tsx
+++ b/app/about/contact/page.tsx
@@ -44,7 +44,7 @@ export default async function ContactUs() {
 
         <div className="content-wrapper flex justify-center px-40">
           <div className="flex flex-wrap justify-center gap-y-6 gap-x-6 xs:w-1/2">
-            {pageContent?.contactUs?.map((card: ContactUsTypes) => (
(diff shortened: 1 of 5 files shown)


app/about/contact/page.tsx
```diff
@@ -44,7 +44,7 @@ export default async function ContactUs() {
 
         <div className="content-wrapper flex justify-center px-40">
           <div className="flex flex-wrap justify-center gap-y-6 gap-x-6 xs:w-1/2">
-            {pageContent?.contactUs?.map((card: ContactUsTypes) => (
+            {pageContent?.contactUs?.cards?.map((card: ContactUsTypes) => (
```
eldu, line 47: Another thing we might want to look at is the typescript definitions in your files. To me it seems like, we shouldn't be needing the `?` checks since it should exist. But ehh another day

eldu (approved):
approving

app/about/contact/page.tsx
ashleysyg: yea i'm really having a time with typescript. i'll make and issue to audit that later

PR #169, opened 2025-07-09
fix: External Links Audit

### LINKED ISSUE #163: chore: external link component audit

ashleysyg:
This PR replaces next links and anchor tags with External Link component

### DIFF
--- a/components/storage/StorageForm.tsx
+++ b/components/storage/StorageForm.tsx
@@ -32,7 +32,7 @@ const Form: React.FC<FormProps> = ({ selectedAnswers, onAnswerChange, questions 
               remarkPlugins={[remarkGfm]}
               components={{
                 a: ({ node, href, ...props }) => (
-                  <ExternalLink href={href || '#'} {...props} />
+                  <ExternalLink href={href!} {...props} />
                 ),
               }}
             >
--- a/components/storage/StorageForm.tsx
+++ b/components/storage/StorageForm.tsx
@@ -4,6 +4,7 @@ import { SelectedAnswers, FormQuestions } from '@/lib/storage-types'
 import Markdown from 'react-markdown';
 import rehypeRaw from 'rehype-raw'
 import remarkGfm from 'remark-gfm'
+import ExternalLink from '@/components/ui/external-link'
 
 interface FormProps {
   selectedAnswers: SelectedAnswers;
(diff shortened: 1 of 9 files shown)


components/storage/StorageForm.tsx
```diff
@@ -30,8 +31,8 @@ const Form: React.FC<FormProps> = ({ selectedAnswers, onAnswerChange, questions
               rehypePlugins={[rehypeRaw]}
               remarkPlugins={[remarkGfm]}
               components={{
-                a: ({ node, ...props }) => (
-                  <a {...props} className="text-keppel-700 hover:text-keppel-500 underline" />
+                a: ({ node, href, ...props }) => (
+                  <ExternalLink href={href || '#'} {...props} className="text-keppel-700 hover:text-keppel-500 underline" />
```
eldu: So this looks like it's either going to an `href` or a `#` ? I think just a `#` doesn't navigate you anywhere though

eldu: unless you're going to a header (id) but that needs like `#section-header`

ashleysyg: This is because of a typescript error. I have to apply my link styling to the a tags in the markdown to override the default markdown styling... but href is a string or undefined for a tags... so the href has to go nowhere if it is undefined. I attempted to clean this up by 1. moving the styling to the external link component here and elsewhere in the code, unless there is custom style overrides. and 2. using href! to assume href will always be defined. do we like that better?

components/ui/external-link.tsx
```diff
@@ -15,13 +15,21 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
   className = '',
   ...props
 }) => {
+  // Default keppel styling for external links
+  const defaultClass = "text-keppel-700 hover:text-keppel-500 underline";
+  const combinedClass = external
+    ? className
+      ? `${defaultClass} ${className}`
+      : defaultClass
+    : className;
```
eldu, line 24:
```suggestion
  const defaultClass = "text-keppel-700 hover:text-keppel-500 underline";
```

components/ui/external-link.tsx
```diff
@@ -15,13 +15,21 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
...
+
   if (external) {
     return (
       <a
         href={href}
         target="_blank"
         rel="noopener noreferrer"
-        className={className}
+        className={combinedClass}
```
eldu, line 32: I think you could just do this

```suggestion
        className={`${defaultClass} ${className || ''}`}
```

components/ui/external-link.tsx
eldu: suggestion to get rid of combined class and instead do it inline below

eldu (changes requested):
Am I not looking at this correctly? This is what is looks like when I run this locally on my computer 
Screenshot 2025-07-14 at 10 03 35 AM

eldu:
closing this one

PR #179, opened 2025-07-11
Feat: adds services/oscar content

ashleysyg:
This PR adds content for services/oscar. It is currently in yaml format... may switch to mdx down the road.


eldu (approved):
yeah i think an mdx for this would be good. another time~ 

approving :)

PR #180, opened 2025-07-11
DELETE Feat: services/stronghold content

### LINKED ISSUE #175: Use MDX for certain pages
There are many pages that could benefit from using MDX files for the content. Such as some of the services pages 

https://nextjs.org/docs/pages/guides/mdx

ashleysyg:
This PR adds stronghold content in mdx format. 

Most of the 119 file changes are moving the content folder into the app folder from the project root. This was necessary for the mdx library to be able to read the content. I then had to make minor changes to other pages besides stronghold to update the path to the content files/folders.

I made the choice to put the page metadata (hero title and description - and eventually buttons as well for those pages that have hero buttons) into the mdx file. This was because I wanted to reduce the number of content files so the content is easier to maintain. Because of that, I ended up having to find esm exports within mdx to put only those metadata into a separate file for use in server components. This happens during build.

I also realized I needed to move as much css/tailwind styles out of the mdx files as possible for non-technical editors. I made wrapper components for common components like Button, SectionHeader, etc. that have the css/tailwind styles in them so the import and use within mdx files is very simple. I generated a readme explaining how to edit content that is in the top of the content folder.


app/services/stronghold/StrongholdMDXClient.tsx
ashleysyg, line 1: i dont know if this is fully necessary. i didn't want services/stronghold/page.tsx to be client side, so i moved the imports to this client side component. it doesn't really matter with the stronghold page because it doesn't have much content... but i thought it might be a good precedent here for other more complex pages that i convert to mdx.

components/ui/mdx-styled-components.tsx
ashleysyg, line 1: added this to try to keep as much code/tailwind out of the mdx file as possible for ease of editing just the content.

lib/extract-mdx-metadata.ts
ashleysyg, line 1: this was a choice made to reduce the number of content files that will all have to be updated and kept in sync. i could have just put the page metadata in a separate metadata.ts file in each content folder, but it seemed worth the extra complexity here to be able to put the metadata into the same mdx file that the content is in

ashleysyg:
I am going to move the generic code changes for mdx to another PR and the stronghold specific changes to its own PR as well.

PR #182, opened 2025-07-14
Fix: External Links audit

ashleysyg:
This PR reverts some of the changes made in the OG external links audit PR, because unnecessary refactoring introduced a bug. I moved default link classes (keppel, underlined, hover makes it lighter keppel) into the external link component because those styles repeat in a few places. However, when I did that, button variant styles were not overriding the external link styles, and it was onerous to debug.

I branched from the first commit of the OG PR, chose a few changed files to commit, but left the breaking changes behind. I  propose deleting the other PR and merging this one to main.


eldu (approved):
great! thanks ashley

PR #183, opened 2025-07-15
Feat mdx content setup

### LINKED ISSUE #175: Use MDX for certain pages

### LINKED ISSUE #177: refactor: move some content to mdx

ashleysyg:
This PR pulls generic mdx code from this one https://github.com/brown-ccv/ccv-website/pull/180/. That PR has Stronghold specific content/code as well. Those changes will go in its own PR that branches off main once this is merged in. Moving forward, new content PRs will likely use mdx as well.

Down the line, more complicated yaml content will be converted to mdx, and the pages code will be simplified!


package.json
ashleysyg, line 1: i think i can uninstall most of these added libraries actually. i believe i'm only using next/mdx

lib/extract-mdx-metadata.ts
ashleysyg, line 1: this might be unecessarily complicated. or be bad practice because it could cause build issues... but i thought it was good to do it this way instead of maintaining a different metadata file for every singel page on the website?

package.json
eldu: did you want to do this as a separate pr to remove them?

lib/extract-mdx-metadata.ts
eldu: i don't think i really understand what these mdx metadata files are?

eldu (approved):
I'm approving this PR to keep the progress going. I responded to your comments also.

Looks like most of this PR is moving into the app folder.

I think in the `tsconfig.json` there is a line for the paths that is commented out. You may need to update the paths there?

eldu (changes requested):
move hero to mdx and remove metadata files / functions to generate it??

package.json
ashleysyg: i did it here

lib/extract-mdx-metadata.ts
ashleysyg: removed!

eldu (approved):
approving! 

in theory i think that the mdx-styled-components could be refactored such that these defaults are in the regular components and only like the props people need to fill out for content are exposed.

PR #184, opened 2025-07-15
chore: google map margins/flex styling improvements

ashleysyg:
This PR keeps the map visible in mobile view and moves it under the 180 george st text in larger windows. It also fixes large screen map view so there are no margins and map takes up the full section.


eldu (approved):
ooohh this looks really good

PR #185, opened 2025-07-15
Style: hero text box

ashleysyg:
This PR adds a right margin to the hero text+buttons div that increases as the window increases. When there is a short description, the text is a single line and the hero looks awkward.


eldu (approved):
LGTM!

PR #194, opened 2025-07-17
fix: slack terms of services

### LINKED ISSUE #193: fix: slack terms of service page down

ashleysyg:
This PR adds the terms of services pdf to content/images, fixing the contact-us/slack button that links to it in a new tab.


ashleysyg:
I don't know why the gpu and matlab images are added in this PR, but they shouldn't be an issue. they were merged into main and should be there.

anna-murphy (approved):
I almost went down a whole rabbit hole looking into NextJS assets and the Page router before I realized that this just works and is fine lmao. Looks great!

PR #196, opened 2025-07-17
Chore remove hero jpeg

ashleysyg:
This PR fixes the warning where hero.jpeg is passed to Hero on many pages. This is from before we changed the Hero from a stock image to a blue gradient.


ashleysyg:
I had to set the base to `feat-content-stronghold-mdx` instead of main because I git switched changes in `feat-content-stronghold-mdx` into this branch accidentally.

PR #200, opened 2025-07-21
style: hero bg

ashleysyg:
This PR adds a navbar blue, purple, and keppel css gradient background to the hero.


anna-murphy (approved):
Wanted to run this to see what it's like but this is cool! 

image

Idk why "and" is wrapping on my screen like that but that sounds like a Different Question 🙃

PR #201, opened 2025-07-21
card style

ashleysyg:
This PR removes shadow and enhances borders to modernize the card design


anna-murphy (approved):
I'm running the code locally and I'm having a hard time like Telling what changed -- the code *looks* fine (it's just CSS anyway) so I'll give it a 👍 but I don't really feel qualified to say if it's a good change or not 😓 -- happy to chat off-github too if that helps!

ashleysyg:
Thanks Anna! The change is very minor - the cards just looked outdated. I removed shadow, and then added a subtle gradient and darker outline to make them stand out from the white background a bit more without the shadow.

PR #210, opened 2025-07-23
feat: create custom 404 page

hetd54:
Uses Nextjs [not found function](https://nextjs.org/docs/app/api-reference/file-conventions/not-found) to serve custom 404 page when function fires

Screenshot 2025-07-23 at 7 16 28 PM

### DIFF
--- a/app/not-found.tsx
+++ b/app/not-found.tsx
@@ -2,15 +2,14 @@ import ExternalLink from "@/components/ui/external-link"
 import { Button } from "@/components/ui/button"
 import React from "react"
 import Hero from "@/components/Hero"
+import Link from "next/link"
 
 export default function NotFound() {
   return (
     <div>
       <Hero title="Oops!" description="It looks like this page does not exist">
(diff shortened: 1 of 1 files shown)


app/not-found.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Button } from "@/components/ui/button"
+import React from "react"
+import Hero from "@/components/Hero"
+
+export default function NotFound() {
+  return (
+    <div>
+      <Hero title="Oops!" description="It looks like this page does not exist">
+        <Button variant="primary_filled" size="xl">
+          <ExternalLink href="/" external={false}>
```
ashleysyg: I would use the regular next <Link/> component for this because that's what I've been using for internal links since we don't need to do the external or not check. Thanks Heather!!

ashleysyg: `<Link href="/">Return Home</ExternalLink>`

ashleysyg (approved):

PR #214, opened 2025-07-25
fix: remove default margins from ccv bars in mdx files

ashleysyg:
This PR removes the y margins from CCV bars so section headers look as intended, without so much separation between the bars and title.

### DIFF
--- a/components/assets/CCVBars.tsx
+++ b/components/assets/CCVBars.tsx
@@ -1,6 +1,5 @@
 import Image from "next/image"
 import CCVBarsImage from "@/assets/CCVBars.png"
-// import { cn } from "@/lib/utils"
 
 export default function CCVBars({ className }: { className?: string }) {
   return (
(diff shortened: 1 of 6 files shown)


components/assets/CCVBars.tsx
```diff
@@ -1,11 +1,10 @@
 import Image from "next/image"
 import CCVBarsImage from "@/assets/CCVBars.png"
-import { cn } from "@/lib/utils"
 
 export default function CCVBars({ className }: { className?: string }) {
   return (
     <Image
-      className={cn("mb-4", className)}
+      className={`m-0 p-0 ${className || ""}`}
```
hetd54: Just tested this, and the globals.css edit covers this, so you don't need m-0 p-0 here :)

hetd54 (approved):
see comment on unneeded styles in ccvbars 👍

components/assets/CCVBars.tsx
ashleysyg: oh nice! thank you. i increased the margin/padding for img tags within mdx and ccv bars!

PR #230, opened 2025-07-30
refactor: PascalCase

hetd54:
Refactored file names, confirmed all exports are PascalCase


ccv-cj:
<details>
<summary>
    <p>Git SHA: <code>59e17e9f804ab0f6319d6b608e283a76ce806aa9</code></p>
</summary>
<pre><code>
{"status": "success", "message": "No accessibility issues found!", "results": []}

</code></pre>
</details>

components/ui/variants.tsx
```diff
@@ -1,6 +1,6 @@
 import { cva, VariantProps } from "class-variance-authority"
 
-export const buttonVariants = cva(
+export const ButtonVariants = cva(
```
ashleysyg, line 3: is your reasoning here keeping variants.tsx lowercase because it's not a component? just curious, i think thats fair.

ashleysyg (commented):
just a question about the case of variants file. maybe we should think about reorganizing the components/ui folder into a more granular/sane grouping (like components/calendar and components/storage currently). but another day, another task.

components/ui/variants.tsx
hetd54: yeah that was my reasoning, but not sure how I feel about it?

PR #446, opened 2025-10-22
fix: make event fetching client side so it fetches on page load

### LINKED ISSUE #359: Events page shows past events
I just noticed that the events page includes past events. Can we please make sure to filter the past events?

gtdang:
Makes the event fetch client-side so that it fetches on page load. Copilot helped a lot on this one because use effects 🤯.

closes #359

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -46,9 +46,7 @@ export default async function Home() {
             <FeaturedCarousel carouselData={featuredCarouselData} />
           </ContentSection>
           <ContentSection title={"Events"} align={"left"} id={"events"}>
-            <Suspense fallback={<Spinner />}>
-              <EventSectionClient />
-            </Suspense>
+            <EventSectionClient />
           </ContentSection>
         </div>
       </>
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -4,9 +4,7 @@ import { HeroCard } from "@/components/card/HeroCard"
 import { ImpactBanner } from "@/components/ImpactBanner"
 import { FeaturedCarousel } from "@/components/carousel/FeaturedCarousel"
 import { CarouselItem } from "@/components/carousel/Carousel"
-import EventSection from "@/components/EventSection"
-import { getEventData } from "@/app/queries"
-import { getStringDate } from "@/components/calendar/utils"
+import EventSectionClient from "@/components/EventSectionClient"
 import React, { Suspense } from "react"
 import Spinner from "@/components/assets/Spinner"
 import { ScrollButton } from "@/components/button/ScrollButton"
@@ -22,15 +20,6 @@ export default async function Home() {
   const featuredCarouselData = featuredCarouselRaw.data.carousel
 
   try {
-    const currentDate = new Date()
-    const today = getStringDate(
-      currentDate.getMonth() + 1,
-      currentDate.getDate(),
-      currentDate.getFullYear()
-    )
-    const futureDates = getEventData(today)
-    const pastDates = getEventData(`-2 months${today}`)
-
     return (
       <>
         <MainHero
@@ -58,12 +47,7 @@ export default async function Home() {
           </ContentSection>
           <ContentSection title={"Events"} align={"left"} id={"events"}>
             <Suspense fallback={<Spinner />}>
-              <EventSection
-                streamedDataPast={pastDates}
-                streamedDataFuture={futureDates}
-                currentDate={currentDate}
-                today={today}
-              />
+              <EventSectionClient />
             </Suspense>
           </ContentSection>
         </div>
--- a/components/EventSection.tsx
+++ b/components/EventSection.tsx
@@ -22,8 +22,8 @@ export interface DataProps {
 }
 
 interface EventSectionProps {
-  streamedDataFuture: Promise<any>
-  streamedDataPast: Promise<any>
+  streamedDataFuture: any[]
+  streamedDataPast: any[]
   today: string
   currentDate: Date
 }
@@ -34,8 +34,8 @@ export function EventSection({
   today,
   currentDate,
 }: EventSectionProps): JSX.Element {
-  const dataFuture = use(streamedDataFuture)
-  const dataPast = use(streamedDataPast)
+  const dataFuture = streamedDataFuture
+  const dataPast = streamedDataPast
 
   return (
     <div className="flex flex-col gap-4 xl:flex-row xl:justify-between xl:gap-24">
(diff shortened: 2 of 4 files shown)


ccv-cj:
<details>
<summary>
    <p>Git SHA: <code>f286c12ce7bb470d08ef84e8a0a00075edc1eeed</code></p>
</summary>
<pre><code>
{
    "total": 17,
    "passes": 17,
    "errors": 0,
    "results": {
        "http://localhost:3000/": [],
        "http://localhost:3000/about/contact": [],
        "http://localhost:3000/services/storage": [],
        "http://localhost:3000/services/storage/compare": [],
        "http://localhost:3000/services/user-support": [],
        "http://localhost:3000/services/virtual-machine-hosting": []
    }
}

</code></pre>
</details>

app/page.tsx
```diff
@@ -58,12 +47,7 @@ export default async function Home() {
           </ContentSection>
           <ContentSection title={"Events"} align={"left"} id={"events"}>
             <Suspense fallback={<Spinner />}>
-              <EventSection
-                streamedDataPast={pastDates}
-                streamedDataFuture={futureDates}
-                currentDate={currentDate}
-                today={today}
-              />
+              <EventSectionClient />
```
anna-murphy: This was all previously in a `<Suspense />` but I think you can remove that now that this is a client component!

components/EventSection.tsx
```diff
@@ -22,8 +22,8 @@ export interface DataProps {
 }
 
 interface EventSectionProps {
-  streamedDataFuture: Promise<any>
-  streamedDataPast: Promise<any>
+  streamedDataFuture: any[]
+  streamedDataPast: any[]
```
anna-murphy: If it was up to me, I'd love to have a better sense of what this type actually is -- maybe not for this PR given that it was previously `Promise<any>`, but if you have bandwidth in the future, sounds like a decent small tech-debt item to handle.

components/EventSectionClient.tsx
```diff
@@ -0,0 +1,47 @@
+"use client"
+
+import React, { useEffect, useState } from "react"
+import { getEventData } from "@/app/queries"
+import { getStringDate } from "@/components/calendar/utils"
+import EventSection from "@/components/EventSection"
+import Spinner from "@/components/assets/Spinner"
+
+export default function EventSectionClient() {
```
anna-murphy, line 8: I like this a lot! There's a part of me that does wonder if this can be modularized, but I think that's something for me to think about later 📝

components/EventSectionClient.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    const now = new Date()
+    setCurrentDate(now)
+    const todayStr = getStringDate(
+      now.getMonth() + 1,
+      now.getDate(),
+      now.getFullYear()
+    )
```
anna-murphy: A small thing for me is that when I'm doing this kind of date manipulation, I tend to get nervous -- everything feels like it's magic numbers and I just don't like it. *If* you feel the same way, I have two suggestions for you. So I have two suggestions for you:
1. Are you using `date-fns`? If so, they probably have a function to do the work you're looking to do.
2. If not, the ISO Date String is a standard spec, so you can do some silly string parsing to get the values you want out of the ISO Date String:

```js
const [date, month, year] = date.toISOString().split("T")[0].split("-");
```

components/EventSectionClient.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Spinner from "@/components/assets/Spinner"
+
+export default function EventSectionClient() {
+  const [futureDates, setFutureDates] = useState<any>(null)
+  const [pastDates, setPastDates] = useState<any>(null)
+  const [currentDate, setCurrentDate] = useState<Date | null>(null)
+  const [today, setToday] = useState<string>("")
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
```
anna-murphy, line 23: This looks good! I think if I was writing this function I would have implemented it _slightly_ different. I don't love defining functions in-line like you did here, even though it's not abnormal in `useEffect`s specifically. Instead, I'd move the function outside the component, and have it return past and future dates that you can set to React state in the promise callback.

Here's kind of what I mean:

```js
async function fetchData(today: string) {
  const future = await getEventData(today)
  const past = await getEventData(`-2 months${today}`)
  return [future, past]
}

// In the component...

useEffect(() => {
  setLoading(true)
  const todayStr = makeTodayStr() // However you want to do this
  fetchData(todayStr).then(([future, past]) => {
    setFutureDates(future)
    setPastDate(past)
    setLoading(false)
  })
}, [])
```

These are logically equivalent, but I think it's easier (for me) to read.

components/EventSectionClient.tsx
```diff
@@ -0,0 +1,47 @@
...
+      now.getMonth() + 1,
+      now.getDate(),
+      now.getFullYear()
+    )
+    setToday(todayStr)
+
+    const fetchData = async () => {
+      setLoading(true)
+      const future = await getEventData(todayStr)
+      const past = await getEventData(`-2 months${todayStr}`)
```
anna-murphy: Fun JS trick! These requests are made in sequence. You fetch one, wait for a response, then fetch a second one and wait for a response. Probably doesn't matter here, but you can speed up your code like this:

```js
const [future, past] = await Promise.all([getEventData(todayStr), getEventData(`-2 months${todayStr}`)])
setFutureDates(future)
// ...
```

This makes your first request, then makes your second request, and then waits for both to return, and then continues execution. Fun!

anna-murphy (commented):
Hi George! @eldu let me know you were waiting on a second pair of eyes on this code, and said I might be able to help! I added a few comments that are mostly stylistic -- your solution looks pretty good to me! You're welcome to take anything here you want if it'll be useful, but don't feel obligated to accept all my comments.

Btw, @eldu mentioned this was because you were noticing that NextJS was fetching during build time and baking the data into the application. We ran into this problem over on the RIDDC project, and I solved it in a slightly different way there: https://github.com/ridatadiscoverycenter/riddc-dashboard/pull/112 You should be able to tell NextJS to force a rerender of the page every page load. For your case, I actually prefer your solution here (this is just one small component on the main page -- you want the main page to be fast), but if you want to have this tool in your toolbox, it might come in handy as well.

gtdang:
Wow! Thanks for all the great input @anna-murphy!

components/EventSectionClient.tsx
gtdang: Oh! It seems that `getStringDate` is just a custom utility to do `date.toISOString().split("T")[0]`. So we can just ❌ all of that!

hetd54 (approved):
ty George!

PR #499, opened 2026-02-09
feat: implement ui frontend for sitewide search

hetd54:
# Search Implementation

## What's Included in the Search Index

The search index contains **static content** from the website's MDX files and referenced data files:

### Indexed Content
- ✅ **Page titles and descriptions** - From MDX frontmatter
- ✅ **Page content** - Full text from MDX files with markdown/JSX stripped
- ✅ **Section headings** - Extracted from `<ContentSection title="...">` components
- ✅ **Team member data** - Names, titles, bios from `content/data/people.yaml`
- ✅ **Project information** - From carousel data files referenced in MDX
- ✅ **Page categories and tags** - From frontmatter metadata

### Index Structure
Each document in the search index looks like this:
```json
{
  "id": "about-us",
  "slug": "/about/us",
  "title": "About Us",
  "description": "The Center for Computation and Visualization...",
  "content": "Full searchable text including team bios...",
  "headings": ["Our Mission", "Our Team", "Our Home"],
  "category": "about",
  "url": "/about/us"
}
```

## What's NOT Included in the Search Index

The search index is **built at deployment time** from static files in the repository. Content that doesn't exist in the `/content` folder at build time cannot be indexed. The following content is **not indexed**:

- ❌ **API Data** - Events and Careers fetched from the Events/Workday API at runtime
- ❌ **Component-Level Information** - Information that is obfuscated within another component

## Meilisearch Architecture

The Meilisearch index is stored in **Google Cloud Run** as part of a containerized Meilisearch instance.

```
┌────────────────────────────────────────────────────────────┐
│  User's Browser                                            │
│                                                            │
│  ┌────────────────────┐                                    │
│  │  Search Component  │                                    │
│  │  (React)           │                                    │
│  └─────────┬──────────┘                                    │
│            │                                               │
│            │ HTTP Request: "search?q=mission"              │
│            │ with MEILISEARCH_SEARCH_KEY                   │
└────────────┼───────────────────────────────────────────────┘
             │
             ↓
┌────────────┼───────────────────────────────────────────────┐
│  Google Cloud Run: Meilisearch                             │
│            │                                               │
│  ┌─────────▼──────────────────────┐                        │
│  │  Meilisearch Engine            │                        │
│  │                                │                        │
│  │  - Receives search query       │                        │
│  │  - Validates API key           │                        │
│  │  - Searches inverted index     │                        │
│  │  - Returns ranked results      │                        │
│  │                                │                        │
│  │  Index Storage (Ephemeral):    │                        │
│  │  ┌──────────────────────────┐  │                        │
│  │  │ Inverted Index           │  │                        │
│  │  │ - Tokenized words        │  │                        │
│  │  │ - Document pointers      │  │                        │
│  │  │ - Ranking metadata       │  │                        │
│  │  │ - Typo tolerance data    │  │                        │
│  │  └──────────────────────────┘  │                        │
│  └────────────────────────────────┘                        │
└────────────────────────────────────────────────────────────┘
```

**Service Details:**
- **Name:** `meilisearch`
- **Region:** `us-central1`
- **Resources:** 1GB RAM, 1 CPU
- **Scaling:** Min 0 instance, Max 1 instance
- **URL:** `https://meilisearch-zvwvypot3q-uc.a.run.app`

### Data Persistence
Currently, the search index is stored **ephemerally** within the Cloud Run instance. This means:
- ✅ Fast access (data in memory/local disk)
- ⚠️ Data persists across deployments if instance stays warm
- ⚠️ Data may be lost if instance is completely replaced

**Advantages:**
- ✅ **Fast search performance** - Data is local to the search engine
- ✅ **Simple architecture** - Single container, no external dependencies
- ✅ **Low cost** - ~$15-30/month for small-medium indexes
- ✅ **Easy deployment** - One workflow file, one container
- ✅ **Built-in scaling** - Cloud Run handles traffic spikes
- ✅ **Acceptable for static content** - Index rebuilds on each deployment anyway

**Disadvantages:**
- ⚠️ **Data loss risk** - If instance completely restarts, index must rebuild
- ⚠️ **No backup/recovery** - Can't rollback to previous index state
- ⚠️ **Rebuild required** - Every deployment runs search indexing

### Alternative: Persistent Storage (Cloud Storage Bucket)

**When to Switch:**
You should consider using Cloud Storage for Meilisearch data if:

1. **Index is large** (>1GB)
   - Static content is unlikely to be this large
   - Dynamic content (events, news) could grow large over time

2. **Rebuilding is expensive**
   - If indexing takes >5 minutes
   - If you're indexing dynamic data from external APIs
   - If you're hitting API rate limits during indexing

3. **You need backups/snapshots**
   - Regulatory requirements for data retention
   - Ability to rollback to previous index state
   - Disaster recovery planning

4. **Frequent updates to dynamic content**
   - Index changes happen outside deployment cycle
   - Want to preserve index between deployments
   - Multiple services write to the same index


ccv-cj:
<details>
<summary>
    <p>Git SHA: <code>58ef18c63e979361abf82b9cab10399d03aa2100</code></p>
</summary>
<pre><code>
{
    "total": 17,
    "passes": 17,
    "errors": 0,
    "results": {
        "http://localhost:3000/": [],
        "http://localhost:3000/about/grant-and-publication-materials": [],
        "http://localhost:3000/about/help": [],
        "http://localhost:3000/services/storage": [],
        "http://localhost:3000/services/virtual-machine-hosting": [],
        "http://localhost:3000/services/workshops": []
    }
}

</code></pre>
</details>

components/DialogSearch.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export function DialogSearch({ searchTitle }: DialogSearchProps): JSX.Element {
+  return (
+    <Dialog>
+      <DialogTrigger asChild>
+        <Button
+          variant="icon_only"
+          iconOnly={<FaMagnifyingGlass />}
+          className="bg-red-university text-white"
```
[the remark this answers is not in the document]
hetd54: We should ask isa what she wants here

components/StyledDialog.tsx
```diff
@@ -28,7 +28,7 @@ const StyledDialogOverlay = React.forwardRef<
   <DialogOverlay
     ref={ref}
     className={cn(
-      "data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 fixed inset-0 z-50 bg-black/80",
+      "data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 fixed inset-0 z-50 bg-black/80 bg-opacity-75",
       className
     )}
```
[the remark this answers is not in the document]
hetd54: tw v3 things

components/Search.tsx
```diff
@@ -0,0 +1,149 @@
...
+      searchClient={searchClient}
+      indexName="pages"
+      insights={false}
+    >
+      <div className="flex flex-col">
+        {/* Search Input */}
+        <div className="relative border-b border-gray-500 px-4 py-3">
+          <SearchBox
+            queryHook={queryHook}
+            placeholder="Search documentation..."
```
eldu: Changing this since this is just site search for right now

```suggestion
            placeholder="Search site..."
```

components/Search.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </div>
+
+        {/* Results */}
+        <SearchResults />
+      </div>
+    </InstantSearch>
+  )
+}
+
+function queryHook(query: any, search: (arg0: any) => void) {
```
eldu: For the types, are the types here really `any`? Are they not strings?

lib/utils.ts
```diff
@@ -71,3 +71,25 @@ export const humanize = (str: string | null | undefined): string => {
   }
   return cleanStr.charAt(0).toUpperCase() + cleanStr.slice(1)
 }
+
+/**
+ * Converts a URL path to breadcrumb format with title case
+ * Example: "/services/ai-tools" -> "Services > Ai Tools"
```
eldu: Ah, I see what's happening here. Would it be worth making a list of words we'd want to capitalize? It would be ideal if `Ai` could be `AI`

lib/utils.ts
```diff
@@ -71,3 +71,25 @@ export const humanize = (str: string | null | undefined): string => {
...
+ */
+export function urlToBreadcrumb(url: string): string {
+  const segments = url
+    .replace(/^\/|\/$/g, "")
+    .split("/")
+    .filter(Boolean)
```
eldu: This can be simplified -- 

`/[^/]+/g`: This looks for any sequence of characters that is not a slash.

One Step: It automatically ignores leading slashes, trailing slashes, and double slashes (//) in one go.

The || []: This is a safety net. If the URL is just /, .match() returns null. This ensures you always get an empty array instead of a crash.

```suggestion
  const segments = url.match(/[^/]+/g) || []
```

lib/utils.ts
```diff
@@ -71,3 +71,25 @@ export const humanize = (str: string | null | undefined): string => {
...
+
+  const formatted = segments.map((segment) => {
+    const cleanSegment = segment.replace(/[_-]/g, " ")
+
+    return cleanSegment
+      .split(" ")
+      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+      .join(" ")
+  })
```
eldu: This can also be simplified (I asked AI since there were multiple slices and joins happening)

Why this is better:
.replace(/[_-]/g, " "): Swaps underscores and dashes for spaces first.

.replace(/\b\w/g, ...): This uses the "word boundary" anchor (\b). It finds the first letter of every word and caps it instantly.

Performance: It processes the string in place rather than creating multiple intermediate arrays via split(" ").

```suggestion
  const formatted = segments.map(segment => 
  segment
    .replace(/[_-]/g, " ")
    .replace(/\b\w/g, char => char.toUpperCase())
);
```

eldu (commented):
Some comments! I'll take another look after the gsdc meeting

components/Search.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return (
+    <div className="max-h-[400px] overflow-y-auto p-2">
+      <Hits
+        hitComponent={Hit}
+        classNames={{
+          root: "",
+          list: "space-y-2",
+          item: "",
```
eldu: Do you need to make the root and item empty? Can they just not be included here?

```suggestion
          list: "space-y-2",
```

eldu (changes requested):
okay i think this looks great! i just have the other comments and questions.

components/Search.tsx
hetd54: Thanks for catching this -- I was messing with different styles and forgot to remove any that I ended up not changing!

lib/utils.ts
```diff
@@ -71,3 +71,32 @@ export const humanize = (str: string | null | undefined): string => {
...
+): string {
+  const segments = url.match(/[^/]+/g) || []
+
+  const formatted = segments.map((segment) => {
+    let result = segment
+      .replace(/[_-]/g, " ")
+      .replace(/\b\w/g, (char) => char.toUpperCase())
+
+    // Capitalize acronyms
+    acronyms.forEach((acronym) => {
```
eldu: This looks like this could be in one step too instead of looping ! And this could be done outside and after the formatted map.

I think this assumes that the acronym is cased like in the url. Which is guess is what your titleCased part is doing. Ri

Right now you have it inside of 2 loops! A map and then a forEach. This logic can be taken out of both

```
// Join acronyms
const acronymPattern = new RegExp(`\\b(${acronyms.join('|')})\\b`, 'gi');
result = result.replace(acronymPattern, (match) => match.toUpperCase());
```

components/Search.tsx
```diff
@@ -0,0 +1,136 @@
...
+            submitIconComponent={() => null}
+            resetIconComponent={() => <X className="h-4 w-4" />}
+            loadingIconComponent={() => null}
+            classNames={{
+              root: "relative",
+              form: "relative flex items-center gap-3",
+              input:
+                "w-full rounded-md mr-6 placeholder:text-gray-500 focus-visible:outline-none focus-visible:ring-ring focus:outline-none focus-visible:ring-2 focus-visible:ring-sunglow-400 [&::-webkit-search-cancel-button]:hidden [&::-webkit-search-decoration]:hidden",
+              submit: "hidden",
+              submitIcon: "",
```
eldu: Can we delete this since it is empty?

eldu (changes requested):
More small changes! Thanks Heather

eldu (approved):
approving !

PR #541, opened 2026-03-27
feat: add documentation to search

hetd54:
This currently doesn't touch the frontend UI


ccv-cj:
<details>
<summary>
    <p>Git SHA: <code>2ee6587382ea81ab601a13dd16276b94043df836</code></p>
</summary>
<pre><code>
{
    "total": 17,
    "passes": 17,
    "errors": 0,
    "results": {
        "http://localhost:3000/": [],
        "http://localhost:3000/about/grant-and-publication-materials": [],
        "http://localhost:3000/about/help": [],
        "http://localhost:3000/services/file-storage-and-transfer": [],
        "http://localhost:3000/services/virtual-machine-hosting": [],
        "http://localhost:3000/services/workshops": []
    }
}

</code></pre>
</details>

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+  )
+
+  if (!Array.isArray(res.data)) return docs
+
+  for (const item of res.data) {
+    if (item.type === "dir") {
+      if (ignoredFolders.includes(item.name)) continue
+      docs.push(
+        ...(await walkRepo(
```
eldu: So, this is a bit of a bottleneck. We have to wait for each file as they process before going to the next file. 

Suggestion to use Promise.all() in order to fire multiple requests simultaneously.

scripts/get-documentation-files.ts
eldu, line 1: So, you might run into Github's rate limits and not actually get all of the files. Consider requiring fewer fetches. Also consider processing like 10 things at a time. So fetching like 10 things at a time before going onto the next batch of 10.

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,176 @@
...
+      owner: OWNER,
+      repo: REPO,
+      path: repoPath,
+      headers: { "X-GitHub-Api-Version": "2022-11-28" },
+    }
+  )
+
+  if (!Array.isArray(res.data)) return docs
+
+  for (const item of res.data) {
```
eldu: The for...of loop is sequential and with the awaits also... It would be great to parallelize !

eldu (changes requested):
Okay so all of my comments are on the `get-documentation-files.ts` file. Basically I think this can be sped up and reduce risk for hitting github's rate limits.

lib/get-open-issues.tsx
```diff
@@ -30,27 +32,23 @@ export async function getOpenIssues() {
...
     allRepos.data.map(async ({ name }: { name: string }) => {
-      const open = await octokit.request(`GET /repos/${org}/${name}/issues`, {
-        org: { org },
-        repo: { name },
+      const open = await octokit.rest.issues.listForRepo({
+        owner: org,
+        repo: name,
         sort: "created",
         direction: "desc",
       })
```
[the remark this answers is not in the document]
hetd54: @eldu Is this an issue? We only have 6 repos, and (ideally) we never really have many open issues (right now it is 0)?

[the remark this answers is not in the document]
eldu: Yeah, let's not paginate the repos or org

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,238 @@
+import { Octokit } from "@octokit/rest"
+import { throttling } from "@octokit/plugin-throttling"
+import { retry } from "@octokit/plugin-retry"
+import { SecretManagerServiceClient } from "@google-cloud/secret-manager"
+import type { SearchDocument } from "@/lib/search-utils"
```
eldu, line 5: This is a comment that is kind of separate from this PR?

My TS Server on my IDE isn't resolving this import. Looks like in the tsconfig.json excludes `scripts` which is the likely cause of this. Any reason why we wanted to exclude scripts?

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,238 @@
...
+        if (!item.sha || !item.path) return []
+
+        const { data, content } = parseFrontmatter(raw)
+        if (data.hidden === "true" || data.hidden === true) return []
+        const chunks = chunkMarkdownByHeadings(content)
+        const pageUrl = buildPageUrl(include.slug, item.path, BASE_URL)
+
+        const fileDocs: SearchDocument[] = []
+
+        for (let i = 0; i < chunks.length; i++) {
```
eldu: The `i` variable here shadows the outer loop's `i`. Suggestion to change this var name. `j` is fine.

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,238 @@
...
+  const variables: Record<string, string> = {
+    owner: OWNER,
+    repo: REPO,
+  }
+
+  paths.forEach((p, i) => {
+    variables[`expression${i}`] = `HEAD:${p}`
+  })
+
+  const result = await octokit.graphql<any>(query, variables)
```
eldu: Can we add a type here?

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,238 @@
...
+  for (let i = 0; i < mdFiles.length; i += batchSize) {
+    const batch = mdFiles.slice(i, i + batchSize)
+    const paths = batch.map((b) => b.path!).filter(Boolean)
+
+    const blobMap = await fetchBlobsGraphQL(octokit, paths)
+    const batchResults = await Promise.all(
+      batch.map(async (item) => {
+        const raw = blobMap[item.path!]
+        if (!raw) return []
+        if (!item.sha || !item.path) return []
```
eldu: Looks like you're filtering mdFiles more but within batch... can we move the filters into the `mdFiles` filter instead?



```suggestion
  const mdFiles = tree.filter((item) => {
    if (item.type !== "blob") return false
    if (!item.path?.toLowerCase().endsWith(".md")) return false
    if (!item.path.startsWith(`${include.slug}/`)) return false
    if (ignoredFiles.includes(path.basename(item.path))) return false
    if (pathHasIgnoredFolder(item.path, ignoredFolders)) return false
    return Boolean(item.sha && item.path)
  })

  const batchSize = 10
  for (let i = 0; i < mdFiles.length; i += batchSize) {
    const batch = mdFiles.slice(i, i + batchSize)
    const paths = batch.map((b) => b.path)

    const blobMap = await fetchBlobsGraphQL(octokit, paths)
    const batchResults = await Promise.all(
      batch.map(async (item) => {
        const raw = blobMap[item.path]
        if (!raw) return []
```

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,238 @@
...
+    auth: token,
+    throttle: {
+      onRateLimit: (retryAfter: any) => {
+        console.warn(`Rate limit hit, retrying after ${retryAfter}s`)
+        return true
+      },
+      onSecondaryRateLimit: (retryAfter: any) => {
+        console.warn(`Secondary rate limit, retrying after ${retryAfter}s`)
+        return true
+      },
```
eldu: Does this retry forever? Maybe we also limit how many times?

eldu (changes requested):
i have some more request changes and questions

scripts/get-documentation-files.ts
eldu: The suggestion highlighting is slightly weird! I was trying to get rid of `        if (!item.sha || !item.path) return []` since that information is available before the blob map unless i'm looking at this weird !

scripts/get-documentation-files.ts
eldu: 3 tries?

scripts/get-documentation-files.ts
hetd54: Looks like this may have been an erroneous addition to tsconfig -- removing,......

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  // fetch tree once
+  const treeRes = await octokit.request(
+    "GET /repos/{owner}/{repo}/git/trees/{tree_sha}",
+    {
+      owner: OWNER,
+      repo: REPO,
+      tree_sha: "HEAD",
+      recursive: "true",
+      headers: { "X-GitHub-Api-Version": "2022-11-28" },
```
eldu: Is there a more recent one we could use?

scripts/get-documentation-files.ts
```diff
@@ -0,0 +1,253 @@
...
+
+// -------------------- Config Fetch --------------------
+async function getDocConfig(octokit: Octokit): Promise<DocConfig> {
+  const res = await octokit.request(
+    "GET /repos/{owner}/{repo}/contents/{path}",
+    {
+      owner: OWNER,
+      repo: REPO,
+      path: "searchConfig.json",
+      headers: { "X-GitHub-Api-Version": "2022-11-28" },
```
eldu: Same question here

eldu (changes requested):

eldu (approved):

PR #575, opened 2026-05-05
feat: Routing from /storage

hetd54:
There was some confusion about the url change from the old site:

>I just found https://ccv.brown.edu/storage (old website) has been moved to https://ccv.brown.edu/services/file-storage-and-transfer?

Thoughts on this redirect? With this PR, folks who visit the old storage page will get redirected to the new page instead of a 404

Per @yangliu2009 :

>Thank you for looking into the issue, Heather. I’d actually recommend not setting up redirects at all. Instead, we can just inform users that the CCV website has migrated and provide the new URLs. This way, we avoid the technical debt of maintaining the old URLs indefinitely. Just my 2c.


paulstey:
I'd be inclined to agree with Yang, and I would vote not to add redirects. That storage page is reachable within two clicks—technically, it's a hover and a click. It's not hard to find!

I'm curious what others think about this.

hetd54:
> I'd be inclined to agree with Yang, and I would vote not to add redirects. That storage page is reachable within two clicks—technically, it's a hover and a click. It's not hard to find!
> 
> I'm curious what others think about this.

I wouldn't mind a top-level redirect (about > about/us, services > services/rates). As you can see, the api routing redirect for that in nextjs is very simple. I agree with not having a redirect for the storage > services/file-storage-and-transfer, though, since that requires an entirely new directory in the project. 

@eldu What do you think?

eldu:
> > I'd be inclined to agree with Yang, and I would vote not to add redirects. That storage page is reachable within two clicks—technically, it's a hover and a click. It's not hard to find!
> > I'm curious what others think about this.
> 
> I wouldn't mind a top-level redirect (about > about/us, services > services/rates). As you can see, the api routing redirect for that in nextjs is very simple. I agree with not having a redirect for the storage > services/file-storage-and-transfer, though, since that requires an entirely new directory in the project.
> 
> @eldu What do you think?

This has been a shift from the old site wherein there were top-level pages for main navbar items (services, about, etc). So those old links for `/about` and `/services` don't exist anymore. 

`about > about/us` seems fine
`services > services/rates` doesn't include all of the other services that we have! 

Let's chat more about this in our website meeting in a little bit heather. 

Maybe we could track down the old links that are used and update those where they are? Maybe once the google indexing is all updated, these old links will also go away from the google search?

hetd54:
Closing, as we aren't redirecting from storage

